PR #26, opened 2021-01-16
Operator Docs

akaszynski:
This PR adds the autogenerated html docs to the sphinx docs and adds to the user guide.

Autogenerated docs from DPF have been patched to match the CSS colors used in sphinx documentation along the simplification and removal of non-CPython content.

image


rlagha (approved):

rlagha:
Very good job Alex, The operators section can only viewed using chrome or Firefox, we may need to add this remark when selecting this section

akaszynski:
Issue with edge is due to the the older version.  We should still add a warning when the page scripts fail to load the operators.  Perhaps implement this on the server side.  I'm hesitant to make too many changes here since they're ephemeral.

PR #29, opened 2021-01-18
Rlagha/example 2

rlagha:


akaszynski:
`00-multi_harmonic.py` is failing because we're missing the `math` operator.  I've patched this in `core.py` by always loading the operator.  I neglected to include this in `core.py`.

`01-modal_cyclic.py` was failing because the spec for `bool_rotate_to_global` has `'B'` instead of `'bool'`.  I've added a check for this.  Ideally it would be included in `mapping_types.py`, but we'd have to refactor the check to allow for a list rather than a 1:1 mapping.  It's easier to patch this with a two line trivial check and then fix this in the server.

I also cleaned up the variable names, spelling, and PEP8 code style.

After running the examples locally, I'm really impressed!  Cyclic results are always a pain and its great to see it implemented in DPF.

akaszynski (approved):

akaszynski:
Just saw that @anslpa was assigned.  Hope you don't mind that I patched the PR.

The docker image needed to be patched as well as I neglected to include `libAns.Dpf.Math.so`, and that might have been hard to fix.  I can give a tutorial about the docker and CI build when I come back from "vacation".

PR #30, opened 2021-01-23
return correction in ...plotter/plot_chart()

ayush-kumar-234:
With this commit and merge request I am testing my repository access rights, please reject the pull request, in case this correction doesn't make sense. 
This is related to the issue reported in DPF-Post repository - [Issue # 15](https://github.com/pyansys/DPF-Post/issues/15).
Thanks!

### DIFF
--- a/ansys/dpf/core/plotter.py
+++ b/ansys/dpf/core/plotter.py
@@ -87,7 +87,7 @@ class Plotter:
         substr = fields_container[0].name.split("_")
         pyplot.ylabel(substr[0] + fieldMin.unit)
         pyplot.title(substr[0] + ": min/max values over time")
-        return pyplot.legend()
+        return pyplot.show()
 
     def plot_contour(self, field_or_fields_container, notebook=None,
                      shell_layers=None, off_screen=None, show_axes=True, **kwargs):


ansys/dpf/core/plotter.py
```diff
@@ -87,7 +87,7 @@ def plot_chart(self, fields_container):
         substr = fields_container[0].name.split("_")
         pyplot.ylabel(substr[0] + fieldMin.unit)
         pyplot.title(substr[0] + ": min/max values over time")
-        return pyplot.legend()
+        return pyplot.show()
```
akaszynski, line 90: Actually, we should refactor this such that we create a `pyplot.figure` and then return the handle to that.  Sometimes (especially with unit testing), having figures pop-up isn't ideal.  If we pass the handle of the figure the user will be able to show the figure at will, and even adjust its size rather than using `pyplot.gcf`

akaszynski:
> With this commit and merge request I am testing my repository access rights

PRs by anyone are permitted, but to merge you'll need to be part of the developers group, the CI tests need to be passing, and you'll need at least one approving review.

ansys/dpf/core/plotter.py
ayush-kumar-234: True, this would be an issue for batch-mode. How about adding a boolean argument to the function which would return the figure handle when set `True`, default value can be `False` which will return `pyplot.show()` and the chart pops-up?

For unit-testing the boolean can be set to `True`? Or when the user wants direct control on the figure handle. This would only add a functionality to the function and the original idea of the function remains intact. As the user expects from the function name `plot_chart` a pop-up 2D chart analogous to `plot_contour`.

akaszynski: > True, this would be an issue for batch-mode. How about adding a boolean argument to the function which would return the figure handle when set `True`, default value can be `False` which will return `pyplot.show()` and the chart pops-up?

Works for me, let's just use `show=True` and then set it to `False` in our unit testing.  Otherwise, let's have it return the figure handle.

ayush-kumar-234: `show=True` sounds good, I will make the necessary changes, thanks!

akaszynski:
@rlagha, when you get a chance, could you please look over this?

PR #35, opened 2021-06-07
Update version 0.3.0

cbellot000:
In this Pull Request all the update from client 0.2.1 to client 0.3.0 are pushed. 
Changes are made:
- issues fixing
- server updates
- new developments


akaszynski:
@cbellot000, you're going to have to update both the docker and windows package (currently stored as a universal artifact) to get this to run correctly.  We've implemented backwards compatibility for older images here in the PyMAPDL repo:
https://github.com/pyansys/PyMAPDL/blob/master/.ci/azure-pipelines.yml

Effectively, we're running two jobs, (one for the older version and one for the newer).

I'll be able to start reviewing this PR once the unit testing is passing for both the old version and the new version of DPF.

ansys/dpf/core/collection.py
```diff
@@ -52,65 +66,103 @@ def set_labels(self, labels):
...
         return self._info['labels']
 
     labels = property(_get_labels, set_labels, "labels")
+    
+    def has_label(self, label):
+        """Check if a collection has a specified label
+
+        Parameters
+        ----------
+        label: string
```
akaszynski, line 126:
```suggestion
        label: str
```

ansys/dpf/core/collection.py
```diff
@@ -52,65 +66,103 @@ def set_labels(self, labels):
...
+            Labels that must be searched, for example ``'time'``.
+            
+        Returns
+        -------
+        bool: True if the specified value have been found in the collection.
```
akaszynski, line 131:
```suggestion
        bool
            ``True`` if the specified value has been found in the collection. 
```

ansys/dpf/core/collection.py
```diff
@@ -52,65 +66,103 @@ def set_labels(self, labels):
...
+
+        Examples
+        --------
+        >>> from ansys.dpf import core as dpf
+        >>> coll = dpf.Collection(dpf.types.field)
+        >>> coll.add_label('time')
+        >>> coll.has_label('time')
+        True
+        >>> coll.has_label('complex')
+        False
```
akaszynski, line 141:
```suggestion

        >>> coll.has_label('complex')
        False
```
Need to have an empty line for this to be rendered correctly in sphinx.

ansys/dpf/core/collection.py
```diff
@@ -52,65 +66,103 @@ def set_labels(self, labels):
...
+        
+        """
+        result = False
+        labels = self.labels
+        i = 0
+        while i < len(labels):
+            if labels[i] == label:
+                result = True
+            i += 1
+        return result
```
akaszynski, line 151:
```suggestion
        return label in self.labels
```
https://stackoverflow.com/questions/7571635/fastest-way-to-check-if-a-value-exists-in-a-list

akaszynski (approved):
Remove `ansys/dpf/__init__.py` and we're good to go.  I'll take another look once merged for a cleanup as this is too big to review effectively.  Unit tests pass and docbuild looks good.

cbellot000:
Hi @akaszynski, is that wanted that we can't create a merge commit but we can only create a squash merge? Since this PR is the result of many others in the other repo, I kept the history when cherry picking in the other repo

PR #42, opened 2021-07-02
update version

cbellot000:


akaszynski (approved):
Totally optional to update the dev version.  It's up to you how you track it.  I'd say if you're going a long time between releases then it's a good idea to increment it.  Otherwise, if you release minor releases often then it's not necessary.

PR #43, opened 2021-07-06
Operator eval method

rlagha:


ansys/dpf/core/dpf_operator.py
```diff
@@ -284,6 +284,32 @@ def __str__(self):
...
+        Parameters
+        ----------
+        pin : int
+
+        Returns
+        --------
+        output : FieldsContainer, Field, MeshedRegion, Scoping, ...
+            by default returns the first output of the operator, and the output of a given pin when specified, or only evaluates the operator without output.
+
+
```
akaszynski, line 300: Ideally there would also be an "Example" section here.  Usually this is the best form of documentation, and should be a fully self contained example using one of the example result files (available from `ansys.dpf.core.examples`).

Something like:
https://github.com/pyansys/pymapdl/blob/2dc08207f6e281c2e3f89aaa7fc1092c4745664c/ansys/mapdl/core/mapdl.py#L171-L179

rlagha: done @akaszynski

akaszynski (approved):
LGTM.  Thanks!

PR #47, opened 2021-07-16
Plotter issue occurs while a fields container contains several fields but that some are empty and ElementalNodal

anslpa:
Plotter issue occurs while a fields container contains several field (that are supposed to be elemental and so plottable) but for some optimization reason, some are still ElementalNodal and empty (as they are empty the process did not change their location while averaging). Then checking the location only when the field is not empty is a solution.

### DIFF
--- a/ansys/dpf/core/plotter.py
+++ b/ansys/dpf/core/plotter.py
@@ -185,11 +185,11 @@ class Plotter:
 
         # pre-loop to get location and component count
         for field in fields_container:
-            #if len(field.data) != 0:
-            location = field.location
-            component_count = field.component_count
-            name = field.name.split("_")[0]
-            break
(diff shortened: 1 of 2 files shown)


akaszynski (approved):
LGTM.

ansys/dpf/core/plotter.py
```diff
@@ -185,11 +185,11 @@ def plot_contour(self, field_or_fields_container, notebook=None,
 
         # pre-loop to get location and component count
         for field in fields_container:
-            #if len(field.data) != 0:
-            location = field.location
-            component_count = field.component_count
-            name = field.name.split("_")[0]
-            break
+            if len(field.data) != 0:
```
akaszynski, line 188: `if len(field.data):` works as well.

PR #51, opened 2021-08-09
Feature: element descriptor

anslpa:
**An element descriptor class is needed to enhance the element description.** 

---------------------

While getting the element type for a given element, an enum value is returned:

`from ansys.dpf import core as dpf`
`from ansys.dpf.core import examples`
`model = dpf.Model(examples.static_rst)`
`elements = model.metadata.meshed_region.elements`
`element = elements[0]`
`type_of_element = element.type`
` type_of_element`
> <element_types.Hex20: 1>

**Issue**: it occurs that only the integer value is returned. 

` type_of_element`
> 1

--------------------
It is now possible to get the element descriptor to know more information about the element.

`element_descriptor = dpf.element_types.descriptor(type_of_element)`
`element_descriptor.name`
> 'hex20'

`element_descriptor.description`
> 'Quadratic 20-nodes Hexa'

`print(element_descriptor)`
> Enum id (dpf.element_types): 1
> Element description: Quadratic 20-nodes Hexa
> Element name (short): hex20
> Element shape: solid
> Number of corner nodes: 8
> Number of mid-side nodes: 12
> Total number of nodes: 20
> Quadratic element: True


ansys/dpf/core/element_descriptor.py
```diff
@@ -0,0 +1,94 @@
...
+        
+        is_beam: bool
+        
+        is_quadratic: bool
+        
+        Examples
+        --------
+        >>> from ansys.dpf import core as dpf
+        >>> element = dpf.ElementDescriptor(10, "Linear 4-nodes Tetrahedron", "tet4", "solid", 4, 0, 4, True, False, False, False)   
+
```
akaszynski, line 50: Class docstring should be placed within the class documentation instead of `__init__` so sphinx can pick up on this.

ansys/dpf/core/element_descriptor.py
```diff
@@ -0,0 +1,94 @@
...
+        txt += str(", ")
+        txt += str(self.number_of_mid_nodes)
+        txt += "\n"
+        txt += "Total number of nodes: "
+        txt += str(self.number_of_nodes)
+        txt += "\n"
+        txt += "Is quadratic element: "
+        txt += str(self.is_quadratic)
+        txt += "\n"
+        return txt
```
akaszynski: Consider using a list rather than adding to a string (since strings are immutable and are copied each time).  It's probably only adding a few microseconds, but it's a better practices to:

```py
lines = []
lines.append("Total number of nodes: ")
lines.append("")

return "\n".join(lines)
```

ansys/dpf/core/element_descriptor.py
```diff
@@ -0,0 +1,94 @@
...
+=================
+"""
+
+class ElementDescriptor: 
+    """Class that describes an element.
+    
+    Examples
+    --------
+    from ansys.dpf import core as dpf
+    element = dpf.ElementDescriptor(10, "Linear 4-nodes Tetrahedron", "tet4", "solid", 4, 0, 4, true, false, false, false)
```
akaszynski, line 12:
```suggestion
    >>> from ansys.dpf import core as dpf
    >>> element = dpf.ElementDescriptor(10, "Linear 4-nodes Tetrahedron", "tet4", "solid", 4, 0, 4, true, false, false, false) 
    >>> element
```

ansys/dpf/core/element_descriptor.py
```diff
@@ -0,0 +1,94 @@
...
+    
+    Examples
+    --------
+    from ansys.dpf import core as dpf
+    element = dpf.ElementDescriptor(10, "Linear 4-nodes Tetrahedron", "tet4", "solid", 4, 0, 4, true, false, false, false)   
+    
+    """
+    
+    def __init__(self, element_id, description, name, shape = None, number_of_corner_nodes = None, number_of_mid_nodes = None, 
+                 number_of_nodes = None, is_solid = None, is_shell = None, is_beam = None, is_quadratic = None):
```
akaszynski, line 17: Remove whitespace around parameters for function signatures.

ansys/dpf/core/element_descriptor.py
```diff
@@ -0,0 +1,94 @@
...
+        >>> from ansys.dpf import core as dpf
+        >>> element = dpf.ElementDescriptor(10, "Linear 4-nodes Tetrahedron", "tet4", "solid", 4, 0, 4, True, False, False, False)   
+        
+        """
+        self.element_id = element_id
+        self.description = description
+        self.name = name
+        self.number_of_corner_nodes = number_of_corner_nodes
+        self.number_of_mid_nodes = number_of_mid_nodes
+        self.number_of_nodes = number_of_nodes
```
akaszynski, line 57: You may consider `self.n_corner_nodes`.

tests/test_elements.py
```diff
@@ -0,0 +1,96 @@
...
+def tetra_element_descriptor(model_elements):
+    el_index = model_elements.scoping.index(3905)
+    return dpf.element_types.descriptor(model_elements[el_index].type)
+
+@pytest.fixture()
+def line_element_descriptor(model_elements):
+    el_index = model_elements.scoping.index(10834)
+    return dpf.element_types.descriptor(model_elements[el_index].type)
+
+def test_hexa_element_descriptor(hexa_element_descriptor):
```
akaszynski, line 33: You may consider using [Parameterized test functions](https://docs.pytest.org/en/6.2.x/parametrize.html) rather than writing out several tests.

ansys/dpf/core/element_descriptor.py
```diff
@@ -64,31 +59,18 @@ def __init__(self, element_id, description, name, shape = None, number_of_corner
...
+        lines.append('-'*18)
+        lines.append("Enum id (dpf.element_types): " + str(self.element_id))
+        lines.append("Element description: " + str(self.description))
+        lines.append("Element name (short): " + str(self.name))
+        lines.append("Element shape: " + str(self.shape))
+        nodes_txt = "Number of corner and mid-side nodes: " + str(self.n_corner_nodes)
+        nodes_txt += str(", ") + str(self.n_mid_nodes)
+        lines.append(nodes_txt)
+        lines.append("Total number of nodes: " + str(self.n_nodes))
+        lines.append("Is quadratic element: " + str(self.is_quadratic))
```
akaszynski, line 73: You may also want to consider f-strings.  For example:

```py
lines.append("Total number of nodes: " + str(self.n_nodes))
```
becomes
```py
lines.append(f"Total number of nodes: {self.n_nodes}")
```
This does preclude its use within Workbench's IronPython env, so I guess we can't if this will be used there as well.

F-strings are really great, and you can even look into correct formatting of the lines to get the printout consistent.  For this, see some of the work here:
https://github.com/pyvista/pyvista/blob/155b932ced36e4fb5a93061e1dd0c73c820e5494/tests/check_example_coverage.py#L157

tests/test_elements.py
```diff
@@ -0,0 +1,81 @@
...
+def line_element_descriptor(model_elements):
+    el_index = model_elements.scoping.index(10834)
+    return dpf.element_types.descriptor(model_elements[el_index].type)
+
+def check_element_attributes(descriptor, element_id, description, 
+                             name, shape, n_corner_nodes, n_mid_nodes, 
+                             n_nodes, is_solid, is_shell, is_beam, 
+                             is_quadratic):
+    assert isinstance(descriptor, dpf.ElementDescriptor)
+    assert descriptor.element_id == element_id
```
rlagha, line 38: enum id?

anslpa:
Discussing with @rlagha . 

__str__ : is quadratic element -> quatric element

anslpa:
> Discussing with @rlagha .
> 
> **str** : is quadratic element -> quatric element

Solved.

PR #55, opened 2021-08-10
Versioning: warning product version

anslpa:
The aim of this PR is to add a warning while using dpf.start_local_server() to mention the consumer version of Ansys product that must be used. 

**Notes about what is already existing:** 
We do already check the DPF server version (since ansys-dpf-core.0.3.0) for few features. Moreover, this check can only be done once we are connected to the server.
We probably would like some backward compatibility that would prevent us to raise any error if the DPF server (directly related to the ansys consumer version) does not match with the python package version.

### DIFF
--- a/ansys/dpf/core/server.py
+++ b/ansys/dpf/core/server.py
@@ -134,6 +134,10 @@ def start_local_server(ip=LOCALHOST, port=DPF_DEFAULT_PORT,
     -------
     server : server.DpfServer
     """
+    from ansys.dpf.core._version import __ansys_consumer_version__
+    import warnings
+    warnings.warn(f"This version of ansys-dpf-core matches with an Ansys {__ansys_consumer_version__} product version.")
+    
     if ansys_path is None:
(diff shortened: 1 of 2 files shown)


ansys/dpf/core/server.py
```diff
@@ -134,6 +134,10 @@ def start_local_server(ip=LOCALHOST, port=DPF_DEFAULT_PORT,
     -------
     server : server.DpfServer
     """
+    from ansys.dpf.core._version import __ansys_consumer_version__
+    import warnings
+    warnings.warn(f"This version of ansys-dpf-core matches with an Ansys {__ansys_consumer_version__} product version.")
```
cbellot000, line 139: @anslpa where does the warning appear at runtime?

cbellot000 (commented):
Could we please add a section in the documentation explaining the versioning relation between the python package and ansys release (0.2.... -> 2021R1 and higher from 2021R2)

PR #56, opened 2021-08-10
Fix: dimensionality misspelling

anslpa:
Dimensionality misspelling fixed.


anslpa:
Ready to be merged. The Python DPF-Post (ansys-dpf-post) module won't be impacted.

ansys/dpf/core/dimensionality.py
```diff
@@ -1,13 +1,13 @@
 """
-Dimensionnality
+Dimensionality
 ===============
```
akaszynski, line 3:
```suggestion
==============
```

ansys/dpf/core/dimensionality.py
```diff
@@ -41,17 +41,17 @@ def __str__(self):
     
     @staticmethod
     def scalar_dim():
-        """Dimensionnality instance corresponding to a scalar field
+        """Dimensionality instance corresponding to a scalar field
 
         Returns
         -------
-        dimensionnality : Dimensionnality
+        dimensionality : Dimensionality
```
akaszynski, line 48: We should consider spelling out the class return types like:

```suggestion
        :class:`ansys.dpf.core.dimensionality.Dimensionality`
```

One of the things we've been doing within `pyaedt` is allowing sphinx to provide links between of return types and class documentation.  It's super helpful.  Also, there's no reason to spell out the parameter name of the return type.

anslpa: Hi @akaszynski, dpf core documentation already provides links between the return type and the considered class (working with this spelling syntax)

rlagha (approved):

PR #68, opened 2021-08-27
Doc/edits starting with model py

PipKat:
Edits go through property field.py


ansys/dpf/core/plotter.py
```diff
@@ -1,6 +1,6 @@
-"""Dpf plotter class is contained in this module.
+"""This module contains the DPF plotter class.
 
-Allows to plot a mesh and a fields container using pyvista.
+This class is used to plot a mesh and a fields container using PyVista.
```
PipKat, line 3:
```suggestion
Contains classes used to plot a mesh and a fields container using PyVista.
```

PR #71, opened 2021-09-22
Add flake8 checks

jleonatti:
This is the first PR to make the code PEP8 compliant. It addresses extra blank lines and other whitespaces issues, covering these flake8 directives: W191, W391, E115, E117, E122, E124, E125, E301. It currently is excluding the generated code under ansys/dpf/core/operators from the check. The formatting of the generated code will be addressed later. Running flake8 was added as part of the style CI.

PR #73, opened 2021-09-22
Remove assert isinstance(cpos, CameraPosition) for plotting tests

anslpa:
This is a requirement to get the CI working following updates in PyVista module. Plotting returns None unless return_camera=True in Plotter.show().

### DIFF
--- a/ansys/dpf/core/field.py
+++ b/ansys/dpf/core/field.py
@@ -253,9 +253,6 @@ class Field(_FieldBase):
         >>> fields_container = disp.outputs.fields_container()
         >>> field = fields_container[0]
         >>> mesh.plot(field)
-        [(1.675707321585373, 1.675707321585373, 2.425707321585373),
-         (0.0, 0.0, 0.75),
-         (0.0, 0.0, 1.0)]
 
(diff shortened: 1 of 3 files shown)


akaszynski:
> @akaszynski Is there anyway to check that the plotting did actually occurred to replace the removed code ?

There is.  I'd write a file to disk with `screenshot='filename.png'` to `tmpdir` and then check that file exists and has a non-zero file size.

Better approach is to add a callback that actually compares the plot with a static known example.  We do that with `pyvista` in https://github.com/pyvista/pyvista/blob/main/tests/plotting/test_plotting.py in:
https://github.com/pyvista/pyvista/blob/9d05d4d0357fbf503975c1c2bcebab7b35455d3b/tests/plotting/test_plotting.py#L89-L154

This may or may not be overkill depending on how much time you want into invest into this.

akaszynski (approved):
I'm happy with this for the moment to get CI passing (I'll remove the version changes on #71).  I recommend in the future writing to file and then testing if the screenshot was generated.

cbellot000 (approved):

PR #77, opened 2021-09-27
Remove unused imports

jleonatti:
Remove unused imports. Add flake8 directive for this check. Add the rest of the standard flake8 directives, except for E501 for line length, which requires some more refactoring.


akaszynski:
Double checking here: are we changing the autogenerated code manually, or changing the autogenerator?

jleonatti:
> Double checking here: are we changing the autogenerated code manually, or changing the autogenerator?

Currently, the autogenerated code is still not being checked with flake8. How I am going to address the style problems with the autogenerated code is still to be determined. I have the idea to run black on the autogenerated code during the generation process, which would save a lot of time trying to update the generator. I would need to discuss this with the dpf team.

rlagha:
> Double checking here: are we changing the autogenerated code manually, or changing the autogenerator?

I confirm that there is no change on the auto-generated code in this PR. @jleonatti

rlagha (approved):

PR #86, opened 2021-10-06
Feat/autogen

jleonatti:
- Add `_build.py` to generate operator Python classes from `_operator_list.py`.
- The `_operator_list.py` file is generated from the operators within `dpf.html`.
- Clean up whitespace in non-autogenerated files within the operators directory.
- A handful of regenerated operator classes are added to this draft PR for review.

Still to be done:
- [ ]  Don't think operator classes are functional as is. Refactor needed there.
- [x] Since operator classes are now pascal case instead of snake case to follow the Python convention, a deprecation warning needs to be added.
- [ ] Incorporate new API to get list of operators instead of using html file.
- [ ] Add readme on how to generate these files.


akaszynski:
If not all the operators are available by scraping the html, we need a reliable way of inspecting or querying the server for the available operators.  @rlagha, do we have a public operator that does this?

ansys/dpf/core/operators/averaging/elemental_difference_fc.py
```diff
@@ -90,36 +106,35 @@ def _spec():
...
-        layers if true (default is false)""",
+                    document="""The max elemental difference is taken through
+        the different shell layers if true
+        (default is false)""",
                 ),
             },
             map_output_pin_spec={
                 0: PinSpecification(
                     name="fields_container",
-                    type_names=['fields_container'],
```
cbellot000: @jleonatti this is needed to work I think

ansys/dpf/core/operators/averaging/elemental_difference_fc.py
```diff
@@ -283,14 +311,13 @@ def fields_container(self):
...
 
         Examples
         --------
         >>> from ansys.dpf import core as dpf
         >>> op = dpf.operators.averaging.elemental_difference_fc()
         >>> # Connect inputs : op.inputs. ...
-        >>> result_fields_container = op.outputs.fields_container() 
-        """
+        >>> result_fields_container = op.outputs.fields_container()
+        """  # noqa: E501
```
cbellot000, line 322: ?

jleonatti:
> If not all the operators are available by scraping the html, we need a reliable way of inspecting or querying the server for the available operators. @rlagha, do we have a public operator that does this?

@cbellot000 has exposed an API to get a list of operators. I have not implemented that here yet.

ansys/dpf/core/operators/averaging/elemental_difference_fc.py
jleonatti: I believe this is fixed now!

jleonatti:
This work is added to the dpf-python-core repo and will be migrated here later.

Issue #12, opened 2020-12-18
Issue trying to get transparency while plotting


anslpa:
While having a shell/solid mesh, the following code added in the plotter.py (plot_contour method, in replacement of plotter.addmesh(...) ) file does not work: 

```
grid["result"] = overall_data
mesh_to_compute = grid.extract_surface()
nan_mask = np.isnan(mesh_to_compute["result"])
val_mask = ~np.isnan(mesh_to_compute["result"])
nan_grid = mesh_to_compute.extract_points(nan_mask, adjacent_cells=False)
val_grid = mesh_to_compute.extract_points(val_mask, adjacent_cells=False)
nan_field = mesh_to_compute["result"][nan_mask]
#nan_field = nan_field[:len(nan_field)-1]
val_field = mesh_to_compute["result"][val_mask]
#val_field = val_field[:len(val_field)-1] ### ISSUE 2 : this must be added in this case to get the plot working
plotter.add_mesh(nan_grid, scalars = nan_field, opacity = 0.3, nan_color = "w", show_edges = False)
plotter.add_mesh(val_grid, scalars = val_field, show_edges = True)
```

Script that must be used to get an issue is: 

```
from ansys.dpf import core

filepath = CORE_PATH + "/tests/rst_operators/multishells.rst"
ds = core.DataSources(filepath)
model = core.Model(filepath) 
mesh = model.metadata.meshed_region
stress = model.operator("SX") ## ISSUE 2 # replace by S
stress.inputs.requested_location.connect('Nodal')
s = stress.outputs.fields_container()

mesh.plot(s[1], shell_layers = core.ShellLayers.TOP)
```

Error seems to be a missing point (on the grid ?):
ValueError: Number of scalars (1198) must match either the number of points (1197) or the number of cells (1692).

PR #120, opened 2021-11-16
Add transparency while plotting a result that is not over the whole mesh

anslpa:
Using the following code, we see that transparency is now working while trying to plot a result not over the whole mesh : 

```
from ansys.dpf import core
core.connect_to_server('10.110.2.65', 50052)
```

```
filepath = examples.multishells_rst
ds = core.DataSources(filepath)
model = core.Model(filepath) 
stress = model.operator("SX") ## ISSUE 2 # replace by S
stress.inputs.requested_location.connect('Nodal')
s = stress.outputs.fields_container()
mesh.plot(s[1], shell_layers = core.shell_layers.top)
```

image

It seems the issue pointed out in #12 is solved. Closing this one.

### DIFF
--- a/ansys/dpf/core/plotter.py
+++ b/ansys/dpf/core/plotter.py
@@ -284,8 +284,8 @@ class Plotter:
 
         val_field = mesh_to_compute["result"][val_mask]
 
-        plotter.add_mesh(nan_grid, scalars = nan_field, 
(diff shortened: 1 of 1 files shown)


anslpa:
This should be enabled in an easier way using features of PR #130.

PR #141, opened 2021-12-08
Plotter: return_cpos (kwargs) updates with a field display

### LINKED ISSUE #140: return_cpos (kwargs) won't work if a field is displayed
Let's assume a model has been defined and we want to plot results. The following command works (no field to display, only mesh):

model_mesh.plot(notebook=False,show_axes=True,return_cpos=True)

The following one does not work:
model_mesh.plot(field_or_fields_container=vmStress.outputs.fields_container(),return_cpos=True)

The same behavior is observed when trying to set cpos:
model_mesh.plot(field_or_fields_container=vmStress.outputs.fields_container(),cpos=[(-100,-100,20),(50,0,0),(0,0,1)])

Here's the  error message I get: 
image

anslpa:
Resolves #140.

### DIFF
--- a/ansys/dpf/core/plotter.py
+++ b/ansys/dpf/core/plotter.py
@@ -280,11 +280,14 @@ class Plotter:
 
         if background is not None:
             plotter.set_background(background)
+            
+        if cpos is not None:
+            plotter.camera_position = cpos
 
         # show result
         if show_axes:
             plotter.add_axes()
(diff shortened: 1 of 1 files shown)


rlagha (approved):

akaszynski:
Note:
Use `Resolves #140.` to automatically link and close out an issue (if that's your goal).

cbellot000 (approved):

PR #95, opened 2021-10-21
Patch 1

PipKat:
This should NOT have been named patch-1. It should have been doc/edits_to_basic_examples. I entered that in the wrong field. Sorry.

Also, I couldn't remember where I left off editing so started back at the basics.

PR #146, opened 2021-12-14
doc/Update basic examples with edits from failed PR #95

PipKat:


PipKat:
@MaxJPRey I think I need some guidance. My PR 95 failed long ago. Alex said that it was because it was done outside of the repo. I tried to submit the same changes in the first two files in two separate PRs. This one failed and I believe the next one will as well. It now says I don't have permissions to write to master and so it forks. How do I edit the PY files for the examples successfully. I must be doing something wrong!

PR #173, opened 2022-02-11
fix release doc with environment variables PowerShell

cbellot000:


.github/workflows/ci.yml
```diff
@@ -288,9 +288,9 @@ jobs:
         run: |
           cd docs/build/html
```
akaszynski, line 289: I’d really recommend using https://github.com/marketplace/actions/deploy-to-github-pages

There’s much less to maintain, code is more succinct, and someone else is improving the functionality and making it cross platform.

rlagha (approved):

.github/workflows/ci.yml
cbellot000: @akaszynski thanks for your input. We added a task to take this comment into account

Issue #186, opened 2022-03-07
[Enhancement] Paths arguments should support pathlib.Path


jfthuong:
All the functions and methods of `pydpf-core` (and more generally `pyansys`) support `str` as when a path is an argument.

Some functions would fail if we provide a `pathlib.Path` (https://docs.python.org/3/library/pathlib.html).

It would be very beneficial to support both types.

The impacted lines are the following:

```
  C:\AnsysDev\Python39\Lib\site-packages\ansys\dpf\core\core.py (24 hits)
	Line 67:     file_path : str
	Line 80:       server_file_path : str
	Line 106:     to_server_folder_path : str
	Line 109:     client_folder_path: str
	Line 136:     server_file_path : str
	Line 139:     to_client_file_path: str
	Line 171:     server_folder_path : str
	Line 174:     to_client_folder_path: str
	Line 205:     file_path : str
	Line 208:     to_server_file_path: str
	Line 221:     server_file_path : str
	Line 241:     path : str
	Line 337:         path : str
	Line 476:         server_file_path : str
	Line 479:         to_client_file_path: str
	Line 520:         server_folder_path : str
	Line 523:         to_client_folder_path: str
	Line 610:         to_server_folder_path : str
	Line 613:         client_folder_path: str
	Line 681:         file_path : str
	Line 684:         to_server_file_path: str
	Line 689:            server_file_path : str
	Line 709:         file_path : str
	Line 722:            server_file_path : str
  C:\AnsysDev\Python39\Lib\site-packages\ansys\dpf\core\data_sources.py (5 hits)
	Line 24:     result_path : str, optional
	Line 71:         filepath : str
	Line 104:         path: str
	Line 133:         filepath : str
	Line 174:         filepath : str
  C:\AnsysDev\Python39\Lib\site-packages\ansys\dpf\core\misc.py (1 hit)
	Line 128:     ansys_path : str
  C:\AnsysDev\Python39\Lib\site-packages\ansys\dpf\core\operators\mesh\stl_export.py (2 hits)
	Line 19:     file_path : str
	Line 166:         my_file_path : str
  C:\AnsysDev\Python39\Lib\site-packages\ansys\dpf\core\operators\metadata\is_cyclic.py (1 hit)
	Line 214:         my_file_path : str
  C:\AnsysDev\Python39\Lib\site-packages\ansys\dpf\core\operators\result\pres_to_field.py (2 hits)
	Line 18:     filepath : str
[...]

PR #191, opened 2022-03-08
Fix/issue#186 path like support

PProfizi:
#186 
Add support for pathlib.Path and other os.PathLike objects as inputs for file paths.

- using in conftest.py:
```
from pathlib import Path
@pytest.fixture(params=[str, Path])
def allkindofcomplexity(request):
    """Resolve the path of the "allKindOfComplexity.rst" result file."""
    return request.param(examples.download_all_kinds_of_complexity())
```
to detect where tests fail due to a Path object being given instead of a string.

- modifying for them all to pass
- reverting the conftest and adding several selected tests where a Path is indeed given as input

### DIFF
--- a/ansys/dpf/core/core.py
+++ b/ansys/dpf/core/core.py
@@ -34,7 +34,7 @@ def load_library(filename, name="", symbol="LoadOperators", server=None):
 
     Parameters
     ----------
-    filename : str
+    filename : str or os.PathLike
         Filename of the operator library.
 
     name : str, optional
@@ -64,7 +64,7 @@ def upload_file_in_tmp_folder(file_path, new_file_name=None, server=None):
 
     Parameters
     ----------
-    file_path : str
+    file_path : str or os.PathLike
         file path on the client side to upload
 
     new_file_name : str, optional
@@ -103,10 +103,10 @@ def upload_files_in_folder(
 
     Parameters
     ----------
-    to_server_folder_path : str
+    to_server_folder_path : str or os.PathLike
         folder path target where will be uploaded files on the server side
 
-    client_folder_path: str
+    client_folder_path: str or os.PathLike
         folder path where the files that must be uploaded are located
         on client side
 
@@ -133,10 +133,10 @@ def download_file(server_file_path, to_client_file_path, server=None):
 
     Parameters
     ----------
-    server_file_path : str
+    server_file_path : str or os.PathLike
         file path to download on the server side
 
-    to_client_file_path: str
+    to_client_file_path: str or os.PathLike
         file path target where the file will be located client side
 
     server : server.DPFServer, optional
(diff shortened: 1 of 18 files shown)


ansys/dpf/core/dpf_operator.py
```diff
@@ -729,9 +729,11 @@ def _fillConnectionRequestMessage(request, inpt, server, pin_out=0):
         workflow,
         time_freq_support,
     )
-
+    from pathlib import Path
```
cbellot000, line 732: pathlib is in the standard python distribution?

cbellot000 (approved):

ansys/dpf/core/core.py
```diff
@@ -517,10 +517,10 @@ def download_files_in_folder(
 
         Parameters
         ----------
-        server_folder_path : str
+        server_folder_path : str or os.PathLike
```
cbellot000, line 520: @PProfizi could you please download the built doc out of the github actions and make sure that the API documentation of what you changed looks good? Because I'm afraid the "or" will not, since we always use ","

ansys/dpf/core/path_utilities.py
```diff
@@ -39,6 +40,9 @@ def join(*args, **kwargs):
     for a in args:
         if isinstance(a, str) and len(a) > 0:
```
cbellot000, line 41: you can make it simpler with:
```
if isinstance(a, (str, Path)) and len(str(a)) > 0:
    parts.append(str(a))
```

ansys/dpf/core/core.py
PProfizi: @cbellot000 
Ok I'll change that. I started using "or" because a lot of times there was already "str, optional" and adding a coma seemed more confusing to me (like: "str, os.PathLike, optional").

ansys/dpf/core/path_utilities.py
PProfizi: sorry, didnt ping you when replying @cbellot000

ansys/dpf/core/core.py
PProfizi: @cbellot000 I found an example of docstring similar to what I did here:

> fields_container (ansys.grpc.dpf.collection_pb2.Collection or FieldsContainer, optional)


in 
> class ansys.dpf.core.custom_fields_container.ElShapeFieldsContainer(fields_container=None, server=None)

ansys/dpf/core/path_utilities.py
jfthuong: or we could do:

```python
if isinstance(a, (str, Path)) and Path(a) != Path(""):
    ...
```

PProfizi: @jfthuong works for me! Modifying now.

Issue #193, opened 2022-03-15
[Question / Enhancement] How to do screenshot of plots?


jfthuong:
2 functions mention `off_screen` argument to be easier for doing a screenshot:
* https://dpfdocs.pyansys.com/api/ansys.dpf.core.plotter.html?highlight=screenshot#ansys.dpf.core.plotter.Plotter.plot_chart
* https://dpfdocs.pyansys.com/api/ansys.dpf.core.meshed_region.html?highlight=screenshot#ansys.dpf.core.meshed_region.MeshedRegion.plot

But the question remains: how to perform a screenshot (i.e. save to a file) a given plot?

I tried using `off_screen=True` and it does not seem to do anything.

With `pymapdl`, you could do something like:
```python
result = pymapdl_reader.read_binary("knight data/Kight_vv.rst")
result.plot_nodal_displacement(0,  screenshot='knight.png')
```

But using the argument `screenshot` does not seem to work in `pydpf`: so how to do?

Ideally, it would be nice to have features like the function `show` of `pyvista`:
https://docs.pyvista.org/examples/02-plot/screenshot.html
... which also gives the possibility to do `plotter.image` to be able to load the picture in matplotlib.

jfthuong:
@PProfizi   Any idea on how to do?  [Ramdane told me to ask you :) ]

PProfizi:
This works for a bare mesh when plotting from a model object:
`model.plot(off_screen=True, screenshot='file.png'`
as this calls directly the pyvista plot function.
I'll look into what happens for plot_contour and MeshedRegion.plot

jfthuong:
I see...

What troubled me is that this does not work in Jupyter (at least in Jupyter run in VS Code).

Maybe we need to add an argument related to `pyvista` or `ipyvtklink` ??

PProfizi:
#194 has been merged.

PProfizi:
@jfthuong The keywords now work, however one has to use the combination:

> model.plot(off_screen=True, notebook=False, screenshot='model_plot.png')

for it to save the image.

It should be working for all `plot` commands.

PProfizi:
Fixing the behavior for certain plot commands when in a Jupyter notebook and using notebook=False with #259.

PProfizi:
Yet, the answer stands:
[...]

PR #194, opened 2022-03-15
Fix plot behavior when using keywords screenshot and off_screen

PProfizi:
Issue #193 
Fixed the behavior to have what would be expected.

Still have to add tests, update the docstrings (some arguments are now kwargs), and update the examples.

PS: I have modified the Plotter class, while work is underway to deprecate this and use DpfPlotter instead. This behavior should be copied when migrating plot_contour to the DpfPlotter.

PPS: when initializing a Plotter directly and then calling plot_contour from it, the off_screen keyword must be given when creating the Plotter as it will be an attribute of this Plotter instance, whereas the screenshot keyword should be given to plot_contour (which is logical but should be explained).


akaszynski:
Rerunning ci by closing/opening

ansys/dpf/core/meshed_region.py
```diff
@@ -406,13 +406,18 @@ def plot(
...
+        pl = _DpfPlotter(self, notebook=notebook, **kwargs)
+        kwargs["screenshot"] = screenshot
+        kwargs["text"] = text
         if field_or_fields_container is not None:
             return pl.plot_contour(
                 field_or_fields_container,
                 notebook,
                 shell_layers,
-                off_screen,
+                # off_screen,
```
anslpa, line 420: @PProfizi loose of compatibility with potential user scripts ? Is it wanted ?

PProfizi:
All plotting has been moved to the new DpfPlotter for 3D plotting.
Still have to modify/improve the plot_chart functionality.

ansys/dpf/core/meshed_region.py
```diff
@@ -406,20 +406,18 @@ def plot(
...
-                show_axes,
-                **kwargs
-            )
+            pl = Plotter(self, **kwargs)
+            return pl.plot_contour(field_or_fields_container, notebook,
+                                   shell_layers, off_screen, show_axes, **kwargs)
 
-        # otherwise, simply plot self
+        # otherwise, simply plot the mesh
+        kwargs["off_screen"] = off_screen
```
cbellot000, line 415: can we not remove that?

cbellot000 (approved):

PR #198, opened 2022-03-17
Fix/timeout server 2tries

PProfizi:
When debugging with PyCharm on my local machine, I kept having a lot of tests fail due to a TimeoutError during Server start-up. It turned out the default timeout of 10s was too little, so I added a second attempt with a timeout of 20s. If it still fails, then it raises the TimeoutError. This did the trick for me.

### DIFF
--- a/ansys/dpf/core/server.py
+++ b/ansys/dpf/core/server.py
@@ -221,15 +221,22 @@ def start_local_server(
 
     server = None
     n_attempts = 10
+    timeout = 10
+    timed_out = False
     for _ in range(n_attempts):
         try:
             server = DpfServer(
-                ansys_path, ip, port, as_global=as_global,
+                ansys_path, ip, port, timeout=timeout, as_global=as_global,
                 load_operators=load_operators, docker_name=docker_name
             )
             break
         except errors.InvalidPortError:  # allow socket in use errors
             port += 1
+        except TimeoutError:
+            if timed_out:
+                break
+            timeout *= 2.
+            timed_out = True
 
     if server is None:
         raise OSError(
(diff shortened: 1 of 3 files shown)


ansys/dpf/core/server.py
```diff
@@ -221,15 +221,22 @@ def start_local_server(
...
+                ansys_path, ip, port, timeout=timeout, as_global=as_global,
                 load_operators=load_operators, docker_name=docker_name
             )
             break
         except errors.InvalidPortError:  # allow socket in use errors
             port += 1
+        except TimeoutError:
+            if timed_out:
+                break
+            timeout *= 2.
```
akaszynski, line 238: After a just a few attempts this timeout would grow quite large. Perhaps consider:
```suggestion
            timeout += 10.
```

cbellot000: @akaszynski in the current implementation, the time out is only increased once

cbellot000 (approved):

cbellot000:
@PProfizi you also told me that the timeout parameter cannot be changed up to the last method using it. Could you please update that as well?

cbellot000:
@PProfizi did you check that when the start_server function is rerun, the first started server (the one we failed ton connect to) is stopped and doesn't leave an unwanted process opened?

ansys/dpf/core/server.py
PProfizi: If it fails for something other than a TimeoutError, it does not go through timeout *= 2, and it can only go through it once. The thing is I did not want to make an assumption on the absolute timeout value given in input (once timeout comes from the calling function) so doubling made more sense than adding *ten something*.

PProfizi:
I'll deal with @cbellot000 's last comments in a new PR. Sorry, I put this one on automatic merge.

Issue #219, opened 2022-04-14
Fix Node labeling in plots


PProfizi:
A bug was found in _InternalPlotter.add_point_labels() regarding node index used fro retrieving the scalar value.

PR #221, opened 2022-04-14
Fixed index issue when getting scalar field value for a node.

PProfizi:
Issue #219, The code used the index from enumerating the nodes instead of using the index_to_id function when retrieving the scalar value associated with a node in Plotter.add_label().


akaszynski:
Let's add a check here.

Also, when you get a chance, it would be great to add code coverage so we can tell if lines or tested or not.

PProfizi:
I could not see how to test this properly as this is really about displaying the right value in the plot.
As for test coverage it is among the things I want to implement in our pipelines as soon as I have time. I'll open an improvement issue.

cbellot000 (approved):

PR #223, opened 2022-04-25
mcMunichBranch

mcMunich:
okay, created a branch of my fork and now am pushing. Will be happy to speak english again afterwards.
In any case, there is a simple cycles to failure example in the basic examples folder that is self explanatory, hopefully.

PR #224, opened 2022-04-25
cycles_to_failure as proposed by McMunich

PProfizi:
Copy of #223 using a branch of a cloned repo so that pipelines can run.


PProfizi:
@mcMunich I have made a PR as contributor so that the pipelines can run.
I just took your example and pushed it via a branch in a cloned repository (and not a forked repository).
This is the way to contribute as of today for PyAnsys projects.
I have not tested it yet.

PProfizi:
Started an Issue with necessary improvements to the pipelines (#225)

mcMunich:
> @mcMunich I have made a PR as contributor so that the pipelines can run. I just took your example and pushed it via a branch of a cloned repository (and not a forked repository). This is the way to contribute as of today for PyAnsys projects. Tested it locally, it works. **I guess personally I would need more explanations as to what is happening to totally understand the result at the end ;)** Right now the pipeline for building the doc fails because pyMAPDL is not installed in our pipelines. TBC

Agreed, I didn't want to overdocument and then have to edit it away. I can add some clarity when I figure out how that works.

PR #229, opened 2022-05-04
Feature: connect_to_server() from a channel

### LINKED ISSUE #200: Allow constructing DpfServer (/ connect_to_server) from a GRPC channel
Note: this issue is mostly copied from the [same issue](https://github.com/pyansys/pymapdl/issues/987) in pydpf, I think a lot of our python libraries share the same problem.

Currently the construction of `DpfServer` only allows to create an unsecure channel from an IP and port.

This prevents taking advantage of many useful features of GRPC:
 - Addressing DPF per host name instead of IP. In a managed environment such as docker-compose or kubernetes, you usually address service by a name like "dpf.svc"
 - Secured TLS channel
 - Interceptors: Interceptors are the center of authentication stacks, and traceability.
 - File-based sockets: GRPC channels can be built from a local file socket, and teams at Ansys have experienced some performance improvements with this approach.

I believe the best way to address this is to allow users to manage their channel, so that all the grpc features are accessible by advanced users:

```py
import grpc
import ansys.dpf.core as dpf
channel = grpc.whatever_construction_method()
channel.some_configuration(...)
server = dpf.connect_to_server(channel=channel)
```

plule-ansys:
Fixes #200 

Add the ability for a user to connect to a running DPF server by providing a channel. This enables advanced scenarios such as DPF running behind a gateway, and a channel with interceptors injecting metadata or credentials. This also allows to address DPF by a DNS name instead of an IP.

This will also be the basis for using DPF with PyPIM (in a further PR)

### DIFF
--- a/ansys/dpf/core/server.py
+++ b/ansys/dpf/core/server.py
@@ -257,7 +257,7 @@ def start_local_server(
     return server
 
 
-def connect_to_server(ip=LOCALHOST, port=DPF_DEFAULT_PORT, as_global=True, timeout=10):
+def connect_to_server(ip=None, port=None, as_global=True, timeout=10, channel=None):
     """Connect to an existing DPF server.
 
     This method sets the global default channel that is then used for the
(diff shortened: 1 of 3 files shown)


plule-ansys:
This will need a different approach to be compatible with future pydpf versions, closing.

Issue #228, opened 2022-04-27
Add wheelhouse to CI/CD


akaszynski:
In support of offline deployment, we need to add a wheelhouse generation and deployment step to our CI/CD. This has been implemented in PyMAPDL:

https://github.com/pyansys/pymapdl/blob/45c89eab352993c3bac237e73e3558974653af84/.github/workflows/ci.yml#L93-L107

This creates *zip artifacts and uploads them on each release step. You'll need to add in additional "smoke tests" for Python 3.7 - 3.9 on Windows and Linux to generate the necessary wheels.

Finally, add or modify a release step to deploy those artifacts automatically for the release. See:

https://github.com/pyansys/pymapdl/blob/45c89eab352993c3bac237e73e3558974653af84/.github/workflows/ci.yml#L318-L361

Specifically:
https://github.com/pyansys/pymapdl/blob/45c89eab352993c3bac237e73e3558974653af84/.github/workflows/ci.yml#L342-L349

PProfizi:
Hey @akaszynski I just started looking into this now I am available, and it seems the wheels are already deployed to Pypi for DPF-Core (see [here](https://pypi.org/project/ansys-dpf-core/0.4.1/#files)).
Is the goal only to get this, and if so is the current method ok, or is it necessary to make our CI more in line with what is done in PyMAPDL?

PProfizi:
@akaszynski see artifacts of example run for wheelhouse generation [here](https://github.com/pyansys/pydpf-core/actions/runs/2295667394)
as well as #230 
As mentioned by @rlagha I will need to check that all wheels in there are in the list of authorized packages, and if not register them.

Is that what is required?

akaszynski:
> Does that completely answer the current need?

Provided the artifacts are also uploaded (automatically) on a release step, it does.

Nice work!

PProfizi:
reopening as questions still need answering before merging:

- package requirements for plotting such as pyvista?
- only create wheelhouses for os/python combinations that have been tested and can be certified?

PProfizi:
Active starting with #351

PR #230, opened 2022-05-09
Ci/release wheels

PProfizi:
#228 
Example of artifacts when running the ci [here](https://github.com/pyansys/pydpf-core/actions/runs/2295667394)


PProfizi:
@rlagha before merging I will need to check that what is contained in the wheelhouses [here](https://github.com/pyansys/pydpf-core/actions/runs/2295667394) is authorized as you mentioned in our discussion.

.github/workflows/ci.yml
```diff
@@ -253,3 +280,244 @@ jobs:
...
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - id: install-dpf
+        uses: pyansys/pydpf-actions/install-dpf-server@v1
+        with:
+          dpf-standalone-TOKEN: ${{secrets.DPF_PIPELINE}}
+          ANSYS_VERSION : ${{env.ANSYS_VERSION}}
+
+      - name: Set AWP_ROOT$env:ANSYS_VERSION
```
cbellot000, line 311: @PProfizi this should already be set in the pydpf-cactions by the lines:
```
   - run: echo "AWP_ROOT${{inputs.ANSYS_VERSION}}=${{env.SERVER}}"  | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
      shell: pwsh
```

.github/workflows/ci.yml
```diff
@@ -253,3 +280,244 @@ jobs:
...
+        shell: cmd
+        run: |
+          pip install -r requirements_build.txt
+          python setup.py bdist_wheel
+          FOR /F %%a in ('dir /s/b dist\*.whl') do SET WHEELPATH=%%a
+          ECHO %WHEELPATH%
+          cd tests
+          pip install %WHEELPATH%
+
+      - name: Test import
```
cbellot000, line 326: @PProfizi there is a lot of duplications between the 2 pipelines, could we not create new reusable actions in the actions repo instead?

.github/workflows/ci.yml
PProfizi: @cbellot000 Ok, noted, I'll remove it and check.

.github/workflows/ci.yml
PProfizi: @cbellot000 Yes there is, and the next step is indeed to work on creating actions. I wanted to finish the pipelines for Linux and for PyDPF-Post before getting those together.

PProfizi:
Irrelevant due to modifications made in parallel PRs.

PR #234, opened 2022-05-10
Ci/linux testing pipeline

PProfizi:
Adding a job for Linux build and test pipeline


PProfizi:
Having random test failures, seems to come from the Linux runner as even steps for publishing to artifacts that worked before now sometimes get stuck. One a test fails, and always for no response from server, all following tests fail too. Seems like a server crash and/or runner issues.

PProfizi:
image


Last run CI, all fail for Python 3.7. Will restrain this PR to adding the Linux job, so will stay on Python 3.8 for now and will start another PR concerning 3.7.
BTW, locally, all tests pass for Windows with Python 3.7 in 1m23s, so the hypothesis of failure due to tests taking too long on the GitHub runners looks most likely. Although the difference between 3.7 and 3.8 is not understood.

PProfizi:
The Windows 3.8 job taking 35min is also not understood yet.
See below for the tests which took longer than 10s (not normal):
image

PProfizi:
Fixed the Windows problem by separating again the tests as was done previously.
Still failing for Linux.

PR #237, opened 2022-05-13
PyPIM integration

plule-ansys:
This PR integrates [PyPIM](https://pypim.docs.pyansys.com/). In an environment configured for PIM, DpfServer is now able to start a remote instance of DPF and connect to it.

The integration approach is intended to follow the current pydpf direction and does not add any gRPC specific dependency. PyPIM itself being based on gRPC, it is lazy loaded under the condition that the user is starting dpf in a pypim environment. It however changes the construction of the channel from "ip/port" to any address as supported by gRPC: https://grpc.github.io/grpc/core/md_doc_naming.html

It currently comes with a limitation that the headers are not transmitted to the instances. It means that this implementation will not work if the DPF instance is behind a gateway, or in an authenticated environment.

### DIFF
--- a/ansys/dpf/core/server.py
+++ b/ansys/dpf/core/server.py
@@ -362,9 +362,10 @@ class DpfServer:
             if is_pypim_configured() and not ansys_path and not docker_name:
                 self._remote_instance = launch_remote_dpf()
                 address = self._remote_instance.services["grpc"].uri
+                # Unset ip and port that may not be the same
+                ip=None
+                port=None
             else:
-                ip = ip or LOCALHOST
-                port = port or DPF_DEFAULT_PORT
                 self._server_id = launch_dpf(str(ansys_path), ip, port,
                                             docker_name=docker_name,
                                             timeout=timeout)
@@ -379,6 +380,8 @@ class DpfServer:
 
         # store the address for later reference
         self._address = address
+        self._ip = ip
+        self._port = port
         self.live = True
         self.ansys_path = str(ansys_path)
         self._own_process = launch_server
--- a/ansys/dpf/core/server.py
+++ b/ansys/dpf/core/server.py
@@ -380,8 +380,8 @@ class DpfServer:
 
         # store the address for later reference
         self._address = address
-        self._ip = ip
-        self._port = port
+        self._input_ip = ip
+        self._input_port = port
         self.live = True
         self.ansys_path = str(ansys_path)
         self._own_process = launch_server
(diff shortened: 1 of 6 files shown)


ansys/dpf/core/server.py
```diff
@@ -351,25 +354,31 @@ def __init__(
...
             raise OSError("DPF does not support Ubuntu")
         elif launch_server:
-            self._server_id = launch_dpf(str(ansys_path), ip, port,
-                                         docker_name=docker_name,
-                                         timeout=timeout)
+            if is_pypim_configured() and not ansys_path and not docker_name:
+                self._remote_instance = launch_remote_dpf()
+                address = self._remote_instance.services["grpc"].uri
+            else:
+                ip = ip or LOCALHOST
```
anslpa, line 366: @plule-ansys I think this must be removed

ansys/dpf/core/server.py
```diff
@@ -351,25 +354,31 @@ def __init__(
...
+        self.channel = grpc.insecure_channel(address)
 
         # assign to global channel when requested
         if as_global:
             dpf.core.SERVER = self
 
         # TODO: add to PIDs ...
 
-        # store port and ip for later reference
-        self._input_ip = ip
```
anslpa, line 371: @plule-ansys would it eventually be possible to keep self._input_ip and self._input_port at None ?

ansys/dpf/core/server.py
```diff
@@ -351,25 +354,31 @@ def __init__(
...
+                address = self._remote_instance.services["grpc"].uri
+            else:
+                ip = ip or LOCALHOST
+                port = port or DPF_DEFAULT_PORT
+                self._server_id = launch_dpf(str(ansys_path), ip, port,
+                                            docker_name=docker_name,
+                                            timeout=timeout)
 
-        self.channel = grpc.insecure_channel("%s:%d" % (ip, port))
+        self.channel = grpc.insecure_channel(address)
```
anslpa, line 372: @plule-ansys see in later packages if the client constructor with full address can be used

ansys/dpf/core/server.py
plule-ansys: oops, done thanks

ansys/dpf/core/server.py
plule-ansys: Sure done. I'm keeping them as-is in the general case, and setting them to "None" when an address is provided.

ansys/dpf/core/server.py
plule-ansys: Yes, hopefully the full address constructor can be used instead of the ip/port...

cbellot000 (approved):

anslpa (approved):

PProfizi:
@plule-ansys Maybe a small test to pass the brand new codecov criteria?

PR #241, opened 2022-05-17
Adding coverage

PProfizi:


MaxJPRey:
LGTM.

MaxJPRey (approved):

codecov.yml
```diff
@@ -0,0 +1,19 @@
...
+    project: false
+    patch:
+      default:
+        # basic
+        target: 50%
+        if_not_found: success
+        if_ci_failed: error
+        if_no_uploads: error
+  ignore:
+    - "examples"  # ignore folder and all its contents
```
cbellot000, line 16: @PProfizi I think we might want to ignore the operator folder which is auto generated as well. It might explain our low coverage percentage

PProfizi: @cbellot000 Indeed it went up to 85%. BTW not sure why we have a conftest.py at the base level and not just the one in tests/

cbellot000: @PProfizi because we use the same conftest to run the doctests

PR #255, opened 2022-06-02
Use pydpf-actions/build_doc to generate doc (adds PDF doc generation)

PProfizi:


PProfizi:
@Revathyvenugopal162 as you might see in the artifacts once the pipelines are done, the pdf has some issues especially at the end of the file.

PR #256, opened 2022-06-02
Improve Compatibility documentation

PProfizi:
- [x] The compatibility table should be at the start of the "Getting Started" page
- [x] It mentions having to synchronize the ansys.grpc.dpf module but not explicitly the server and python client compatibility
- [x] The "Architecture" section mentions Ansys 2021R2 as required while the "Installation with PIP" section mentions "Ansys 2021 R1 or later" and gives only the command pip install ansys-dpf-core which is wrong is using Ansys 2021R1 (should be pip install ansys-dpf-core<0.3.0)
- [x] We might actually want to mention compatibility in the ReadMe, along with specifying that the installation command is for Ansys 2021 R2 onward.
- [x] We can also customize the Issue template to add a tick box with "I made sure I am following the Compatibility guidelines", with an hyperlink to the doc.
- [ ] The Ansys 2021R1 row of the table should be visually separated in some way (thicker boundaries?) --> Could not find how to do this

### DIFF
--- a/docs/source/getting_started/index.rst
+++ b/docs/source/getting_started/index.rst
@@ -4,6 +4,46 @@
 Getting Started
 ===============
 
+Compatibility
+~~~~~~~~~~~~~
+DPF supports Windows 10 and CentOS 7 and later.  For
+more information, see `Ansys Platform Support <https://www.ansys.com/solutions/solutions-by-role/it-professionals/platform-support>`_.
+
+Other platforms may be supported by using DPF within a
+containerization ecosystem such as Docker or Kubernetes.
+For more information, see :ref:`docker`.
+
+***************************
+Client Server Compatibility
+***************************
+
+The `ansys.grpc.dpf <https://pypi.org/project/ansys-grpc-dpf/>`_ module should be synchronized
+with the server version as shown here:
+
+.. list-table:: Client-Server Compatibility
+   :widths: 35 35 35
+   :header-rows: 1
+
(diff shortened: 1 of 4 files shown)


docs/source/getting_started/index.rst
```diff
@@ -4,6 +4,49 @@
 Getting Started
 ===============
 
+Compatibility
+~~~~~~~~~~~~~
+DPF supports Windows 10 and CentOS 7 and later.  For
```
cbellot000, line 9:
```suggestion
DPF supports Windows 10 and CentOS 7 and later. For
```
(spacing)

docs/source/getting_started/index.rst
```diff
@@ -4,6 +4,49 @@
...
+more information, see `Ansys Platform Support <https://www.ansys.com/solutions/solutions-by-role/it-professionals/platform-support>`_.
+
+Other platforms may be supported by using DPF within a
+containerization ecosystem such as Docker or Kubernetes.
+For more information, see :ref:`docker`.
```
cbellot000, line 14: @PProfizi I don't think that the docker page is up to date, we should not add links to it until it is, what do you think?

cbellot000 (approved):

docs/source/getting_started/index.rst
PProfizi: @cbellot000 agreed.

PR #259, opened 2022-06-09
Fixes the use of notebook argument for plot commands 

PProfizi:
When in a Jupyter notebook, the notebook=False argument was not working when using a plot command going through Plotter.plot_contour

### DIFF
--- a/ansys/dpf/core/plotter.py
+++ b/ansys/dpf/core/plotter.py
@@ -712,31 +712,32 @@ class Plotter:
             overall_data[ind] = field.data[mask]
 
         # create the plotter and add the meshes
-        background = kwargs.pop("background", None)
-        cpos = kwargs.pop("cpos", None)
-        return_cpos = kwargs.pop("return_cpos", None)
 
         # add meshes
         kwargs.setdefault("show_edges", True)
         kwargs.setdefault("nan_color", "grey")
         kwargs.setdefault("stitle", name)
+
         text = kwargs.pop('text', None)
         if text is not None:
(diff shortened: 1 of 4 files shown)


ansys/dpf/core/plotter.py
```diff
@@ -770,7 +749,7 @@ def plot_contour(
             version_to_reach = '0.32.0'
```
anslpa, line 749: Hi @PProfizi, could you please check that it does not break back the following : https://github.com/pyansys/pydpf-core/issues/140 ?

ansys/dpf/core/plotter.py
```diff
@@ -601,10 +601,7 @@ def plot_chart(fields_container, off_screen=False, screenshot=None):
     def plot_contour(
             self,
             field_or_fields_container,
-            notebook=None,
             shell_layers=None,
-            off_screen=None,
-            show_axes=True,
```
anslpa, line 607: @PProfizi Would it be possible to check that the show_axes (not used in any examples I think) is still working ?

anslpa (approved):

ansys/dpf/core/plotter.py
PProfizi: It is in the new example where all previous arguments are shown explicitly, with their default PyVista values.

ansys/dpf/core/plotter.py
PProfizi: @anslpa I did, I found a bug, I fixed it, it should work :)

PR #262, opened 2022-06-09
Ci/add concurrency-add back feature template

PProfizi:
Adding concurrency to workflow runs so that a commit to a PR will run the associated workflows while cancelling the ones still running from the previous commit. Same for pushes to tagged branches.

Also adding back the feature issue template. Apparently having a custom bug.yml meant the organization's template for feature was not used either.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,10 @@ on:
       - master
       - "release*"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 env:
   PYANSYS_OFF_SCREEN: True
   DPF_PORT: 32772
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -11,6 +11,10 @@ on:
       - master
       - "release*"
 
+concurrency:
```
cbellot000, line 14: @PProfizi what do that mean?

PProfizi: @cbellot000 
> When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending. Any previously pending job or workflow in the concurrency group will be canceled. To also cancel any currently running job or workflow in the same concurrency group, specify cancel-in-progress: true.

see [here](https://docs.github.com/en/actions/using-jobs/using-concurrency)
It means committing to your PR will cancel an already running workflow for the same PR.
When you work like me and make frequent commits, you often get the same workflow running multiple times in parallel, once for every commit, instead of just for the latest commit.

PR #263, opened 2022-06-14
Fix support for show_axes and scalar bar title for plot_contour

PProfizi:


anslpa (approved):
It would be great to gather the duplicated code as previously discussed.

PR #264, opened 2022-06-14
Add field unit to scalar bar title

PProfizi:


ansys/dpf/core/plotter.py
```diff
@@ -131,7 +131,8 @@ def add_field(self, field, meshed_region=None, show_max=False, show_min=False,
                   label_text_size=30, label_point_size=20, **kwargs):
         # Get the field name
         name = field.name.split("_")[0]
-        kwargs.setdefault("stitle", name)
+        unit = field.unit
+        kwargs.setdefault("stitle", f"{name} ({unit})")
```
anslpa, line 135: @PProfizi shouldn't you use the kwargs = self._set_scalar_bar_title(kwargs) method ?

ansys/dpf/core/plotter.py
```diff
@@ -705,7 +707,7 @@ def plot_contour(
         kwargs.setdefault("nan_color", "grey")
 
         # Set the scalar bar title
-        kwargs.setdefault("stitle", name)
+        kwargs.setdefault("stitle", f"{name} ({unit})")
```
anslpa, line 710: Same here ?

PProfizi: @anslpa same here, this defines the default stitle, then it is set in the following line

anslpa (approved):

PR #269, opened 2022-06-29
Feat/plot warped mesh

PProfizi:
This PR adds the possibility to use a result to warp the mesh with.
This result should give a 3D vector field.

This enables for example the plotting of deformed geometries by giving as input a displacement result at a specific time step:
`disp_result = model.results.displacement.on_time_scoping([step])`
`mesh.plot(warping_field=disp_result, scaling_factor=scaling_factor)`


ansys/dpf/core/field.py
```diff
@@ -262,12 +264,20 @@ def plot(self, shell_layers=None, **kwargs):
         shell_layers : shell_layers, optional
             Enum used to set the shell layers if the model to plot
             contains shell elements. The default is ``None``.
+        scaling_result : result operator, optional
```
cbellot000, line 267: This class name will reflect poorly in the doc, the real type name needs to be used.
I also think that restricting the input type to "Result" is not a good idea, Operators (maybe even Field and Fields Containers) should be supported. "Result" is a helper but not a main DPF entity

ansys/dpf/core/plotter.py
```diff
@@ -376,10 +418,15 @@ def add_mesh(self, meshed_region, **kwargs):
...
-        self._internal_plotter.add_mesh(meshed_region=meshed_region, **kwargs)
+        self._internal_plotter.add_mesh(meshed_region=meshed_region,
+                                        scaling_result=scaling_result,
+                                        scaling_factor=scaling_factor,
+                                        **kwargs)
 
     def add_field(self, field, meshed_region=None, show_max=False, show_min=False,
-                  label_text_size=30, label_point_size=20, **kwargs):
+                  label_text_size=30, label_point_size=20,
+                  scaling_result=None, scaling_factor=1.0,
```
cbellot000, line 428: I don't understand why we have scaling_result both for add_mesh, add_field and show_figure

examples/05-plotting/05-plot_on_warped_mesh.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Plot it on the deformed geometry directly
+disp_field.plot(scaling_result=disp_result, scaling_factor=scaling_factor,
+                title='Field', text='Field.plot()')
+# or by applying it to the mesh
+mesh.plot(disp_field, scaling_result=disp_result, scaling_factor=scaling_factor,
+          title='MeshedRegion', text='MeshedRegion.plot(disp_field)')
+
+# Split the model by material and plot the deformed MeshesContainer obtained
+split_mesh_op = dpf.Operator("split_mesh")
```
cbellot000, line 43: old operator syntax

setup.py
```diff
@@ -5,7 +5,7 @@
 
 from setuptools import setup
 
-install_requires = ["psutil", "progressbar2", "numpy", "ansys.grpc.dpf>=0.2.3"]
+install_requires = ["psutil", "progressbar2", "numpy", "ansys.grpc.dpf>=0.2.3", "protobuf<=3.20.1"]
```
cbellot000, line 8: to be merged?

ansys/dpf/core/plotter.py
PProfizi: @cbellot000 It is in add_mesh and add_field because when we plot a bare mesh, only add_mesh is called and I still wanted to give the possibility to deform it.
As for show_figure, it is not in the arguments.
It is however in Plotter.plot_contour

cbellot000: makes sense, sorry for the show_figure, it's because of how github show difference on large files

setup.py
PProfizi: @cbellot000 Indeed it has been merged in the meantime. A merge with the current master has been added.

PR #273, opened 2022-06-30
Update extras require.

akaszynski:
Loosen extras require and require a newer pyvista.

### DIFF
--- a/ansys/dpf/core/plotter.py
+++ b/ansys/dpf/core/plotter.py
@@ -57,7 +57,7 @@ class _PyVistaPlotter:
     """The _InternalPlotter class is based on PyVista."""
     def __init__(self, **kwargs):
         # Import pyvista
-        from vtk_helper import PyVistaImportError
+        from ansys.dpf.core.vtk_helper import PyVistaImportError
         try:
             import pyvista as pv
         except ModuleNotFoundError:


cbellot000 (approved):

PProfizi:
@akaszynski does this mean we can loose the VTK9 test in vtk_helper.py [here](https://github.com/pyansys/pydpf-core/blob/92f0eaa2a191d1fd3015714f1e48e41e154521a3/ansys/dpf/core/vtk_helper.py#L210)?

PProfizi:
- [ ] Requiring pyvista >= 0.32.0 also means we can drop all checks for the pyvista version in plotter.py concerning return_cpos

akaszynski:
> @akaszynski does this mean we can loose the VTK9 test in vtk_helper.py [here](https://github.com/pyansys/pydpf-core/blob/92f0eaa2a191d1fd3015714f1e48e41e154521a3/ansys/dpf/core/vtk_helper.py#L210)?

Not quite. You can still use VTK v8 with pyvista. If you wanted to make VTK v9 a strict requirement you could then drop support. That's your call, but I think given that Python 3.7 doesn't even have a VTK v9 wheel, you should keep supporting it (until we drop python 3.7 support).

See https://github.com/pyvista/pyvista/issues/2867

PR #298, opened 2022-07-08
Release/test new release workflow

PProfizi:


.ci/build_doc.bat
```diff
@@ -1,7 +0,0 @@
-set SPHINX_APIDOC_OPTIONS=inherited-members
```
cbellot000, line 1: why are we deleting this? In the release 0.5.dev2 I updated this file and the changes are important to remove many docs warnings

cbellot000: yes, thanks @PProfizi

PProfizi:
Deprecated

PR #300, opened 2022-07-11
Prepare pipelines for v222 and fix doctests

PProfizi:


setup.py
```diff
@@ -5,7 +5,7 @@
 
 from setuptools import setup
 
-install_requires = ["psutil", "tqdm", "numpy", "ansys-dpf-gate", "protobuf<=3.20.1"]
+install_requires = ["psutil", "tqdm", "numpy", "ansys-dpf-gate==0.1.dev1", "protobuf<=3.20.1"]
```
anslpa, line 8: Why is this needed ?

PR #309, opened 2022-07-15
Migrate from setup.py to pyproject.toml

jorgepiloto:
Partially solves for https://github.com/pyansys/pydpf-core/issues/296.


jorgepiloto:
It is a bit tricky to track the actions log when they call the https://github.com/pyansys/pydpf-actions/ ones.

jorgepiloto:
Having private discussions to see which is the best approach to face previous issue.

PR #317, opened 2022-07-20
Fix doc for release 0.5.dev2

cbellot000:


PProfizi:
@cbellot000 shouldn't we change this?
image

cbellot000:
> @cbellot000 shouldn't we change this? image

you're right, updated

PR #341, opened 2022-07-26
Add testing on Linux for ANSYSR222

PProfizi:


tests/conftest.py
```diff
@@ -21,6 +21,8 @@
 ACCEPTABLE_FAILURE_RATE = 0
 
 core.settings.disable_off_screen_rendering()
```
cbellot000, line 23: @PProfizi why do you need the environment varibales? core.settings.disable_off_screen_rendering() serves the same purpose

PProfizi: @cbellot000 I know, but for Linux it seems it did not work, it kept freezing at the doctest for plotter.show_figure

cbellot000 (approved):

PR #342, opened 2022-07-26
add Pipeline running examples with/without gatebin

cbellot000:
Add new pipeline under GithubActions -> run examples without and without gatebon for the 3 different server types


.ci/run_non_regression_examples.py
```diff
@@ -0,0 +1,55 @@
...
+DPF_SERVER_TYPE = os.environ.get("DPF_SERVER_TYPE", None)
+if DPF_SERVER_TYPE:
+    if DPF_SERVER_TYPE == "INPROCESS":
+        core.SERVER_CONFIGURATION = core.AvailableServerConfigs.InProcessServer
+    elif DPF_SERVER_TYPE == "GRPC":
+        core.SERVER_CONFIGURATION = core.AvailableServerConfigs.GrpcServer
+    elif DPF_SERVER_TYPE == "LEGACYGRPC":
+        core.SERVER_CONFIGURATION = core.AvailableServerConfigs.LegacyGrpcServer
+
+list_tests = [
```
cbellot000, line 23: @anslpa, @PProfizi I chose these tests as "non-regression" test. Does that seem good to you?

anslpa: @cbellot000 It looks good to me.

anslpa (approved):

anslpa:
@cbellot000 Could you please point out what will need to be done to run the example "with gatebin" once this is available ?

cbellot000:
> ebin into the .ci/yml file and then call the following steps with each server typ

@anslpa I was sthinking that gatebin would be installed by default in our actions, and we will just uninstall it to test without

PR #350, opened 2022-07-27
Fix circular ref between mesh and nodes and elements

cbellot000:


ansys/dpf/core/collection.py
```diff
@@ -41,11 +41,6 @@ class Collection:
         default is ``None``, in which case an attempt is made to use the global
         server.
 
-    Examples
```
PProfizi, line 44: @cbellot000 Why was this one problematic?

ansys/dpf/core/meshed_region.py
```diff
@@ -133,9 +133,7 @@ def elements(self):
         DPF Elements object with 8 elements
 
         """
-        if self._elements is None:
-            self._elements = Elements(self)
-        return self._elements
```
PProfizi, line 138: @cbellot000 That was a quick fix for a problem with vtk grids.

ansys/dpf/core/collection.py
cbellot000: @PProfizi it's not a supported behavior we should use the implementation instead (MeshesContainer here)

ansys/dpf/core/meshed_region.py
cbellot000: @PProfizi what do you mean? I didn't take the fix from anywhere, I'm fixing a circular ref here

PProfizi: @cbellot000 ah yes sorry I confused this with another part of code.

PProfizi (approved):

anslpa (approved):

PR #351, opened 2022-07-27
Add wheelhouse generation for CI

PProfizi:
Activate the option for wheelhouse creation when running the normal CI.
Right now this creates a wheelhouse only for Windows 3.8, without plotting dependencies such as pyvista and matplotlib.
Wheelhouse tested in a local venv and working.


PProfizi:
@cbellot000 @anslpa I tested the wheelhouse locally in a venv and the import works.

PProfizi:
image

cbellot000 (approved):

PR #352, opened 2022-07-27
Use DPF_SERVER_TYPE environment variable to switch the default server type

anslpa:
They are currently 3 server types that can be used in pydpf-core. The default one can now be set using the DPF_SERVER_TYPE environment variable. 

The following script can be used to start an InProcess DPF server:
> import os
> os.environ["DPF_SERVER_TYPE"] = "INPROCESS"
> dpf.start_local_server()


anslpa:
Unit tests for v222 and v221 have been run locally.

tests/test_launcher.py
```diff
@@ -192,3 +194,47 @@ def test_server_ip(server_type_remote_process):
...
+    try_serv_type = os.environ.get(dpf_server_type_str, None)
+    if try_serv_type:
+        raise Exception(
+            "Fixture is not correctly working"
+        )  # a specific case is already set to run the unit tests
+    else:
+        if SERVERS_VERSION_GREATER_THAN_OR_EQUAL_TO_4_0:
+            # test for v222 and higher
+            os.environ[dpf_server_type_str] = "GRPC"
+            my_serv = core.start_local_server(as_global=False)
```
PProfizi, line 214: @anslpa Are we sure all servers will correctly close?

PProfizi (approved):

tests/test_launcher.py
anslpa: I added "shutdown()" calls for the started servers.

cbellot000 (approved):

ansys/dpf/core/server_factory.py
```diff
@@ -6,7 +6,10 @@
 protocols and server configurations available.
```
anslpa, line 6: We need to think about a way of not starting many servers while being in process (maybe it could be managed in that feature)

PR #364, opened 2022-07-29
Documentation: remove generated files and rename one example folder that it is taken into account

anslpa:
Averaging examples are now '08' labelled. 
Remove generated files (examples 06 and 07).


anslpa:
image

Documentation ok

PR #375, opened 2022-08-01
Example/average across bodies

vsousacosta:
Add example 08-averaging/01-average_across_bodies


PProfizi:
I get weird connectivity issues when looking at the doc
image

examples/08-averaging/01-average_across_bodies.py
```diff
@@ -89,7 +97,10 @@ def average_across_bodies(analysis):
     min_max.inputs.fields_container.connect(stresses)
     max_val = min_max.outputs.field_max()
 
-    mesh.plot(stresses)
+    # Uncomment this block to obtain the plot
```
PProfizi, line 100: @vsousacosta Does this mean the plotting function does not work correctly?

vsousacosta: Actually, the plotting function works well. The problem apparently has something to do with the document builder, as the plotting issue only happens when building the docs (it works normally when running the code). So, in order to avoid it, I commented the mesh.plot() functions and inserted screenshots of the images I needed.

PProfizi: Still, this means there is a bug when generating the plots for the doc, which should be the way to do it. This might have something to do with the package versions used (Pyvista, vtk, etc...) when building the doc vs when running it locally, or with the backend used.

examples/08-averaging/01-average_across_bodies.py
```diff
@@ -0,0 +1,283 @@
...
+
+def average_across_bodies(analysis):
+    # This function will extract the stresses in the Z direction (with the average
+    # across bodies property activated) and plot them.
+
+    # Create a model from the simulation results.
+    model = dpf.Model(analysis)
+    mesh = model.metadata.meshed_region
+
+    # We're interested in the last time step, so:
```
cbellot000, line 103: @vsousacosta this is what we call "time set" and not "time step" if you can rename it please.
Time step is another concept, a step as different substep and usually you have apply different loads on different steps

examples/08-averaging/01-average_across_bodies.py
```diff
@@ -0,0 +1,283 @@
...
+    time_step = 3
+
+    # Extracting the stresses in the Z direction. By default, DPF already applies
+    # averaging across bodies when extracting the stresses.
+    stress_op = ops.result.stress_Z()
+    stress_op.inputs.connect(model)
+    stress_op.inputs.time_scoping.connect(time_step)
+    stress_op.inputs.requested_location.connect("Nodal")
```
cbellot000, line 111: could you please use:
```suggestion
    stress_op.inputs.requested_location.connect(dpf.locations.nodal)
```

cbellot000 (approved):

PR #376, opened 2022-08-01
Fix version of latest Ansys unified install available in grpc compatibility error message

PProfizi:


ansys/dpf/core/server_types.py
```diff
@@ -277,7 +277,7 @@ def check_ansys_grpc_dpf_version(server, timeout):
                             f"from Ansys {ansys_version_to_use})"
                             f" and the ansys-grpc-dpf version installed ({grpc_module_version})."
                             f" Please consider using the latest DPF server available in the "
-                            f"2022R1 Ansys unified install.\n"
+                            f"2022R2 Ansys unified install.\n"
```
anslpa, line 280: @PProfizi can you use the string value of server_to_ansys_version in _version.py ?

anslpa: I think {ansys_version_to_use} is correctly set for that usage

PProfizi: @anslpa I think it was a requirement, for the message to mention the latest ANSYS unified install.
{ansys_version_to_use} is the one to use, but not necessarily the latest available.

cbellot000 (approved):

PR #389, opened 2022-08-08
Add a manually triggered workflow for testing of pydpf-post

PProfizi:
The goal is to first add a manually triggered workflow for testing PyDPF-Post.
The manually triggered action offers to select the Core branch on which to run the workflow.
The PyDPF-Post branch name to test against can be given as input and defaults to "master".
This is to test the workflow and later integrate it in regular testing.
This needs to be merged in master before it can be tested.
See:
https://levelup.gitconnected.com/how-to-manually-trigger-a-github-actions-workflow-4712542f1960

### DIFF
--- a/.github/workflows/pydpf-post.yml
+++ b/.github/workflows/pydpf-post.yml
@@ -1,0 +1,78 @@
+name: PyDPF-Post testing
+
+on:
+  workflow_dispatch:
+    inputs:
+      PyDPF-Post_branch:
+        description: 'PyDPF-Post branch to test'
+        required: true
+        default: 'master'
+        type: string
+
+env:
+  PACKAGE_NAME: ansys-dpf-core
+  MODULE: core
+  ANSYS_VERSION: 222
+
+jobs:
+  Clone_and_Test:
+    name: Clone and Test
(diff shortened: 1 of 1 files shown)


.github/workflows/pydpf-post.yml
```diff
@@ -0,0 +1,78 @@
+name: PyDPF-Post testing
```
cbellot000, line 1: @PProfizi could we not use the action already written in the post? It looks like the same thing.
Using something like:

      - name: "Test"
        uses: pyansys/pydpf-post/test-post@v1.0
        with:
          python-version: ${{ matrix.python-version }}
          ANSYS_VERSION: ${{env.ANSYS_VERSION}}
          PACKAGE_NAME: ${{env.PACKAGE_NAME}}
          PyDPF-Post_branch: 'master'

PProfizi: @cbellot000 I wanted to minimize the use of external actions first for easier debugging, but yes a composite action could definitely be used, although there is none in the Post repo right now.

cbellot000 (approved):

PR #403, opened 2022-08-12
Add explicit testing of all supported Python versions

PProfizi:
Fixes support for vtk>=9.1.0.
Makes Python 3.10 available.
There is still a problem with Python 3.10 on Ubuntu-18.04, see #416.

### DIFF
--- a/.github/workflows/retro.yml
+++ b/.github/workflows/retro.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
         os: ["windows-latest", "ubuntu-18.04"]
     env:
       ANSYS_VERSION: 221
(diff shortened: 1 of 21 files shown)


PProfizi:
R222 Python3.7 Ubuntu-18.04: 
**test_python_plugins.py** tests [fail](https://github.com/pyansys/pydpf-core/runs/7802826633?check_suite_focus=true#step:7:1095) due to hashlib import, apparently because of it not finding openssl.
A problem with the provided python environment for **test_python_plugins**?

@cbellot000 Strangely this is a problem only for Python 3.7 on Ubuntu-18.04.

PProfizi:
Run_examples Python3.* Ubuntu-18.04:
The **Install OpenGL** step must be changed to the pyvista/setup-headless-display action to be compatible with Linux.

.github/workflows/retro.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
cbellot000, line 23: @PProfizi could you please either make a doc or setup a meeting (maybe with a slide on which pipelines are available) to decide when each pipeline is ran?

rlagha (approved):

PR #426, opened 2022-08-24
Round 1 of standardizing DPF-Core doc

PipKat:


PipKat:
@Roberto I meant to start with DPF-Post doc but accidentally started with DFP core. This doesn't address Vale stuff but just tries to reorganize the doc content to be more like other repos. They use a "docs" directory rather than a "doc" directory. Should this be changed for consistency sake? When I changed the name of the "APIs" section to "API reference", a local doc generation didn't show the APIs. Is this expected? I didn't submit a change to the api/index.rst file, just in case it would mess up the doc for this section.

RobPasMue (approved):
LGTM! It's a good restructuring of the docs. Thanks @PipKat.

PProfizi:
@PipKat I've realized there is a problem with doc generation, going back a few days actually. I have to investigate but in the meantime I'd rather not merge a PR concerning the doc.
If you do not mind maybe PyDPF-Post would be a better candidate. We will try not to merge anything here until this issue is resolved.

PR #427, opened 2022-08-25
implemented review suggestions

ayush-kumar-423:
implemented review suggestions


ayush-kumar-423:
@akaszynski the review suggestions are implemented, not sure why the pipeline fails here. Can you please check?

PProfizi:
Hey @ayush-kumar-423 , I don't know if you noticed but the doc generated by the pipeline is not actually not there. 
There is a problem currently with the doc generation, despite the job saying it passes. (see [here](https://github.com/pyansys/pydpf-core/pull/426#issuecomment-1226841125))
I am investigating what the problem is (see #429 )

PProfizi (approved):

PR #472, opened 2022-09-08
Fix/doc building

PProfizi:
Solves https://github.com/pyansys/pydpf-core/issues/428

The Documentation check passes but the documentation artifact is actually empty (easily verified with its small size of 48 kB, should be more than 100MB).
The last known Documentation generation job to work was [here](https://github.com/pyansys/pydpf-core/actions/runs/2901998839) on August 22.
Was rerun to check for unseen change in a dependency, it still passed.

Seems to have started with bump of Graphviz to 0.20.1 [here](https://github.com/pyansys/pydpf-core/actions/runs/2754279313). Cannot find an occurrence before that.

Two problems here:

the documentation check did not see something was wrong with the doc (artifact too small / step too short)
no error was raised although something went obviously wrong


PProfizi:
Solved in #472

PR #493, opened 2022-09-21
Feat/system coupling mapping

PProfizi:


.ci/build_doc.bat
```diff
@@ -3,7 +3,7 @@ call sphinx-apidoc -o ../docs/source/api ../ansys ../ansys/dpf/core/log.py ^
 ../ansys/dpf/core/help.py ../ansys/dpf/core/mapping_types.py ../ansys/dpf/core/ipconfig.py ^
 ../ansys/dpf/core/field_base.py ../ansys/dpf/core/cache.py ../ansys/dpf/core/misc.py ^
 ../ansys/dpf/core/check_version.py ../ansys/dpf/core/operators/build.py ../ansys/dpf/core/operators/specification.py ^
-../ansys/dpf/core/vtk_helper.py ^
+../ansys/dpf/core/vtk_helper.py ../ansys/dpf/core/operators/mapping/create_sc_mapping_workflow.py ^
```
cbellot000, line 6: @PProfizi why do you remove that from the doc?

PProfizi: @cbellot000 Sorry, I understood the operator was not meant to be in the general doc. I reverted this. Shouldn't we, however, indicate in some way that this operator is only available with a specific plugin?

cbellot000 (approved):

Issue #498, opened 2022-09-22
Improve TypeHints print in the doc


PProfizi:
### Description of the feature

Following [this pyansys guideline](https://dev.docs.pyansys.com/doc-style/docstrings.html#type-hints) would improve the layout of our API documentation by removing typehinting from the signature and instead putting it at the level of each parameter.
This also removes the need to write the type required at the dosctring level apparently. 

### Steps for implementing the feature

https://dev.docs.pyansys.com/doc-style/docstrings.html#type-hints

### Useful links and references

https://dev.docs.pyansys.com/doc-style/docstrings.html#type-hints

PR #500, opened 2022-09-22
Improve TypeHints

GuillemBarroso:
Solves #498. Adding extensions in the proper order as mentioned in [dev guide](https://dev.docs.pyansys.com/doc-style/docstrings.html#type-hints).

Note that I have maintained the order of all other extensions, adding two before and after sphinx.ext.napoleon.

If this is correct I will proceed to do the same in pydpf-post (see pyansys/pydpf-post#160).


PProfizi:
Hi @GuillemBarroso, just to let you know that despite the check being green for the doc generation job, if you take a look at the artifacts you will see that the doc only is 22kB, which means it has not been generated.
This is a current problem with the doc generation job, adressed in #491 , yet for now you should take a look at the doc job to see what went wrong.

GuillemBarroso:
Thank you for the heads up @PProfizi. 

I have just checked the artifacts in pyansys/pdf-post#164 and it looks like the documentation was properly generated. It should have something to do with the order of the extensions. I'll try to fix it now.

PProfizi (approved):
Checked the doc, LGTM. Wil lhave to play around with this feature more.
Especially, I think we do not really see it at work since most types are already declared in the docstring rather than with (or in addition to) typehinting in the signature.

PR #511, opened 2022-09-27
Enable plots of ElementalNodal data

### LINKED ISSUE #373: Implement plotting of ElementNodal
As of now, `ElementNodal` plotting [isn't enabled](https://dpfdocs.pyansys.com/user_guide/plotting.html), but it's something that could be added with [pyvista #3112](https://github.com/pyvista/pyvista/pull/3112). If you prefer not waiting for the next release, instead of `DataSet.separate_cells` simply use `shrink(1.0`).

Here's a brief example:

```py
import ansys.dpf.core as dpf
model = dpf.Model(examples.simple_bar)

nodal_forces = model.results.element_nodal_forces()
fields = nodal_forces.outputs.fields_container()
field = fields.select_component(0)[0]

# separate individual cells
grid = mesh.grid.shrink(1.0)

scalars = field.data
grid.plot(scalars=scalars)
```

tmp

PProfizi:
Resolves #373

### DIFF
--- a/ansys/dpf/core/plotter.py
+++ b/ansys/dpf/core/plotter.py
@@ -697,9 +697,11 @@ class Plotter:
             mesh_location = mesh.nodes
         elif location == locations.elemental:
             mesh_location = mesh.elements
+        elif location == locations.elemental_nodal:
+            mesh_location = mesh.elements
         else:
             raise ValueError(
-                "Only elemental or nodal location are supported for plotting."
+                "Only elemental, elemental nodal or nodal location are supported for plotting."
             )
 
         # pre-loop: check if shell layers for each field, if yes, set the shell layers
--- a/ansys/dpf/core/plotter.py
+++ b/ansys/dpf/core/plotter.py
@@ -712,7 +712,9 @@ class Plotter:
             if shell_layer_check in [
                 eshell_layers.topbottom,
                 eshell_layers.topbottommid,
-            ] and not location == locations.elemental_nodal:
+            ]:
+                if location == locations.elemental_nodal:
+                    print("trying to plot ElementalNodal values for shells.")
                 changeOp.inputs.fields_container.connect(fields_container)
                 sl = eshell_layers.top
                 if shell_layers is not None:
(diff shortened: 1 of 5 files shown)


PProfizi:
For quadratic elements, PyVista's shrinked grid expects a certain number of data values while there is less available in the DPF Field's data array.
size(Field.data) < size(grid.active_scalar)

ansys/dpf/core/plotter.py
```diff
@@ -693,13 +693,22 @@ def plot_contour(
...
         if location == locations.nodal:
             mesh_location = mesh.nodes
         elif location == locations.elemental:
             mesh_location = mesh.elements
+        elif location == locations.elemental_nodal:
+            mesh_location = mesh.elements
         else:
             raise ValueError(
-                "Only elemental or nodal location are supported for plotting."
+                "Only elemental, elemental nodal or nodal location are supported for plotting."
```
PipKat, line 711:
```suggestion
                "Only elemental, elemental nodal, or nodal location are supported for plotting."
```
We use the Oxford comma, which is to always place a comma before the "and" or the "or" in a list with three or or more items.

src/ansys/dpf/core/plotter.py
```diff
@@ -317,8 +317,16 @@ def add_field(
...
+        elif location == locations.elemental_nodal:
+            mesh_location = meshed_region.elements
+            # If ElementalNodal, first extend results to mid-nodes
+            field = dpf.core.operators.averaging.extend_to_mid_nodes(
+                field=field
+            ).eval()
         else:
-            raise ValueError("Only elemental, nodal or faces location are supported for plotting.")
+            raise ValueError(
+                "Only elemental, elemental nodal, nodal or faces location are supported for plotting."
```
PProfizi, line 328:
```suggestion
                "Only elemental, elemental nodal, nodal, faces, or overall location are supported for plotting."
```

PProfizi:
Hey @rafacanton it looks like something was fixed for 25R2 that allows this new feature to work.
Not sure if I should try to find a workaround for older DPF versions, or if I should mark the option as only available starting with 25R2.
It may also be specific to this file. I think I remember something linked to the number of entities reported in result fields for shells vs the number of elements reported in the mesh.

rafacanton (approved):

PR #513, opened 2022-09-27
Update MeshedRegion.grid coordinates when necessary and clear data before plotting.

PProfizi:
to allow for plotting of modified meshes with mesh setters.

Use case:

```
import numpy as np

model = dpf.Model(examples.simple_bar)
ncoord_f = model.metadata.meshed_region.nodes.coordinates_field
model.metadata.meshed_region.plot()

my_field_rotation_matrix = dpf.fields_factory.create_scalar_field(1)
my_field_rotation_matrix.data = np.array([[1],[0],[0],[0],[0],[1],[0],[-1],[0]])

op = dpf.operators.geo.rotate()
op.inputs.field.connect(ncoord_f)
op.inputs.field_rotation_matrix.connect(my_field_rotation_matrix)

op = dpf.operators.geo.rotate(field=ncoord_f, field_rotation_matrix=my_field_rotation_matrix)

result_field = op.outputs.field()

mesh.nodes.coordinates_field = result_field
mesh.plot()
```

- [x] Update the vtk grid associated to the mesh when the coordinates are modified via `set_coordinates_field` or `set_property_field(coordinates)` (and by extension via `nodes.coordinates`). -> this uses the `pyvista.UnstructuredGrid.points` setter.
- [x] Update the vtk grid associated to the mesh when the connectivity is modified via `set_property_field( nodal_properties.nodal_connectivity or element_properties.connectivity)`. -> there is no setter for `pyvista.UnstructuredGrid.cells` or `pyvista.UnstructuredGrid.cell_connectivity`, so the only solution seems to invalidate the grid. This requires more work, yet to solve the current reported bugs this is not necessary as they only concern updating the coordinates.

- [x] Also reset the grid's data using `grid.clear_data()` when using `Plotter.plot_contour` or `DpfPlotter.add_mesh`.


ansys/dpf/core/meshed_region.py
```diff
@@ -285,6 +285,7 @@ def set_coordinates_field(self, coordinates_field):
         coordinates_field : PropertyField or Field
         """
         self._api.meshed_region_set_coordinates_field(self, coordinates_field)
+        self._full_grid = None
```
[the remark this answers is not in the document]
PProfizi: @rlagha FYI I modified the proposed code as requested.

ansys/dpf/core/meshed_region.py
```diff
@@ -285,6 +285,10 @@ def set_coordinates_field(self, coordinates_field):
         coordinates_field : PropertyField or Field
         """
         self._api.meshed_region_set_coordinates_field(self, coordinates_field)
+        if self._full_grid is not None:
```
cbellot000, line 288: @PProfizi In my opinion, we need to think larger. Here the "real" issue is not "when set coordinates is called the grid needs to be updated" but "when the mesh is updated (equivalent to setters are called) the grid is out dated".
With your we will have the same bug if set connectivity, or set element types is called

ansys/dpf/core/plotter.py
```diff
@@ -391,6 +391,8 @@ def add_mesh(self, meshed_region, deform_by=None, scale_factor=1.0, **kwargs):
         >>> pl.add_mesh(mesh)
 
         """
+        if meshed_region.grid is not None:
```
cbellot000, line 394: Have you checked what happens with meshes containers?

PProfizi: It works.

ansys/dpf/core/meshed_region.py
```diff
@@ -207,6 +214,7 @@ def _get_unit(self):
         """
         return self._api.meshed_region_get_unit(self)
 
+    @clear_grid
```
PProfizi, line 217: @cbellot000 Now that I think about it, do we actually want to reset the grid when changing the unit? Unless we add a mechanism to scale the mesh when the unit is changed, there is actually no benefit to this as I have not found any notion of unit in PyVista's UnstructuredGrid.
```suggestion
```

cbellot000 (approved):

PR #545, opened 2022-10-11
Fix 03-create_entities example

kmahajan-cadfem:
- fixed element_ids, previously it was only 1 for all elements
- changed naming from id -> n_id and e_id

PR #550, opened 2022-10-12
Fix 03-create_entities example

PProfizi:
Mirror of #545 which is a PR from a forked repository, hence cannot run the pipelines.
This PR is used to pass the checks, and allow the other PR to be merged without the requirements.


PProfizi:
@kmahajan-cadfem FYI

Issue #466, opened 2022-09-05
``upload_file_in_tmp_folder`` doesn't seem to work on docker image


germa89:
### Before submitting the issue

- [X] I have checked for [Compatibility](https://dpfdocs.pyansys.com/getting_started/index.html#compatibility) issues
- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

It seems that the method ``upload_file_in_tmp_folder`` does not work.

I keep getting an error on the script detailed below.

### Steps To Reproduce

First start the docker image using:

```pwsh
docker run --name dpfserver --restart always -v `pwd`:/dpf -p 50058:50054 ghcr.io/pyansys/dpf-core/dpf:v2021.1 > logdpf.txt
```
Copy the following RST to your working directory:

[file.zip](https://github.com/pyansys/pydpf-core/files/9490673/file.zip)

Then run the following script:
```py
import os

from ansys.dpf import core as dpf


rst_path = os.path.join(os.getcwd(), "file.rst")
dpf.connect_to_server(port=50058)

server_file_path = dpf.upload_file_in_tmp_folder(rst_path)
model = dpf.Model(server_file_path)
print(model)
```

The error should be:

```py

>>> import os
>>> from ansys.dpf import core as dpf
>>> rst_path = os.path.join(os.getcwd(), "file.rst")
>>> dpf.connect_to_server(port=50058)
>>> server_file_path = dpf.upload_file_in_tmp_folder(rst_path)
>>> model = dpf.Model(server_file_path)
>>> print(model)
---------------------------------------------------------------------------
DPFServerException                        Traceback (most recent call last)
Input In [2], in <cell line: 11>()
      9 server_file_path = dpf.upload_file_in_tmp_folder(rst_path)
     10 model = dpf.Model(server_file_path)
---> 11 print(model)

File C:\ProgramData\Miniconda3\envs\dpfcorereleased\lib\site-packages\ansys\dpf\core\model.py:199, in Model.__str__(self)
    197 txt = "DPF Model\n"
    198 txt += "-" * 30 + "\n"
--> 199 txt += str(self.results)
    200 txt += "-" * 30 + "\n"
    201 txt += str(self.metadata.meshed_region)

File C:\ProgramData\Miniconda3\envs\dpfcorereleased\lib\site-packages\ansys\dpf\core\model.py:155, in Model.results(self)
    111 """Available results of the model.
    112
    113 Organizes the results from DPF into accessible methods. All the available
[...]

PR #553, opened 2022-10-12
Fix upload_file_in_tmp_folder

PProfizi:
Pipelines have been seen failing due to `ansys.dpf.core.core.upload_file_in_tmp_folder` returning an empty string, making for subsequent failed calls to `load_library` and such.
This PR aims at investigating why the returned string is sometimes empty by adding a test on the obtained result to the call to `self._api.data_processing_upload_file`.
This has also been a source of [random failing examples](https://github.com/pyansys/pydpf-core/actions/runs/3234208216/jobs/5297030793#step:10:30) during the documentation generation.

Edit: the idea of adding a check for empty paths returned by upload_file_to_tmp is good. Yet the source of the problem might be:
- a lack of TMP env variable (which should be set and does not explain the randomness of the failure)
- the server having not finished to properly start, hence the request would go wrong? That would at least explain the randomness.


germa89:
Isn't this related to #466  ??

PProfizi:
> @PProfizi isn't this related to #466 ??
> 
> By the way, I re-run your tests, it seems the problem was a 503 error. Now it works.

Hi @germa89, the part about the empty string seems to be the same thing, yes. Yet the `DPFServerException: a unimplemented error ocurred:` error is not the same, so that might be specific to Docker.

PProfizi:
Closing as stale. The errors mentioned seems to have been solved in the meantime.

PR #570, opened 2022-10-18
Fix FieldsContainer.animate for specific cases. 

PProfizi:
Fix the animation of a multi-component field without a "deform_by" argument.
Raise an error if fields_container obtained from the evaluation of a result or operator "deform_by" argument is not of the same length as the FieldsContainer being animated.
Improve example.


cbellot000 (approved):
If I understood correctly the test_animator have no assert ? They only check that it didn't throw?

PProfizi:
> If I understood correctly the test_animator have no assert ? They only check that it didn't throw?

Some basic ones do when checking simple things, but otherwise no, as animating a FieldsContainer does not return anything, except for a potential GIF.

PR #470, opened 2022-09-06
Reorganize testing

PProfizi:
This PR reorganizes tests according to a testing matrix defined by the core developer team.
The goal is to spread different test requirements at different steps of the development process to both increase development ease and ensure a thorough robust testing.

 The testing configuration to implement is:
 
On commit to a draft PR:
- [x] unit testing on the latest standalone, for Python 3.7, on windows-latest and ubuntu-latest.

On "ready for review", on commit to a non-draft PR, or on push to master, thus **required to pass for merging on master**:
- [x] unit testing, on the latest standalone, for Python 3.7, on windows-latest and ubuntu-latest.
- [x] examples, on the latest standalone, for Python 3.7, on windows-latest and ubuntu-latest.
- [x] compatibility with the latest PyDPF-Post, on the latest standalone, for Python 3.7, on windows-latest and ubuntu-latest.
- [x] unit testing, on **every** older standalone, for Python 3.7, on windows-latest and ubuntu-latest.
- [x] documentation, on the latest standalone, for Python 3.7, on windows-latest.

On push of a release tag, thus **required to pass for releasing**:
- [x] unit testing, on the latest standalone, for all supported Python versions, on windows-latest and ubuntu-latest.
- [x] examples, on the latest standalone, for all supported Python versions, on windows-latest and ubuntu-latest.
- [x] compatibility with the latest PyDPF-Post, on the latest standalone, for Python 3.7, on windows-latest and ubuntu-latest.
- [x] unit testing, on **every** older standalone, for Python 3.7, on windows-latest and ubuntu-latest.
- [x] compatibility with the latest PyDPF-Post,  on **every** older standalone, for Python 3.7, on windows-latest and ubuntu-latest.
- [x] documentation, on the latest standalone, for Python 3.7, on windows-latest.
- [x] gatebin test, on the latest standalone, for Python 3.7, on windows-latest and ubuntu-latest.

On a weekly basis, run automatically:
- [x] The same as for a release (without the part for drafting a release).

We also give the capability to run manually, with choice of standalone, OS, and Python versions to test:
- [x] Unit tests.
- [x] Examples.
- [x] Compatibility with the latest PyDPF-Post
- [x] Documentation
- [x] Gatebin test.

PR #574, opened 2022-10-19
Update deprecated "set-output" command in CI

PProfizi:
[Reference](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)


PProfizi:
Done in #470

PR #582, opened 2022-10-25
Fix deformation in fields container animate

PProfizi:
An error has been found in the computation of the deformation field.

### DIFF
--- a/ansys/dpf/core/fields_container.py
+++ b/ansys/dpf/core/fields_container.py
@@ -489,7 +489,7 @@ class FieldsContainer(Collection):
         """
         return self.get_label_scoping("time")
 
-    def animate(self, save_as=None, deform_by=None, scale_factor=1.0, mesh=None, **kwargs):
+    def animate(self, save_as=None, deform_by=None, scale_factor=1.0, **kwargs):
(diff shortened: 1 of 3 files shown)


anslpa (approved):

ansys/dpf/core/fields_container.py
```diff
@@ -517,6 +517,11 @@ def animate(self, save_as=None, deform_by=None, scale_factor=1.0, **kwargs):
         # Define the field extraction using the fields_container and indices
         extract_field_op = dpf.core.operators.utility.extract_field(self)
 
+        loop_over = self.get_time_scoping()
+        frequencies = self.time_freq_support.time_frequencies
```
cbellot000, line 521: @PProfizi I don't understand this changes, it means that we can't animate if we don't have a time freq support?

PProfizi: @cbellot000 That was already the case, I just moved this up. Right now the animation is based on the TimeFreqSupport, yes.

cbellot000: to add in the US to handle having no time freq support and to be able to create an animator with a mesh + a fields container

PR #585, opened 2022-10-25
Fix fields container animate key for Linux

PProfizi:
Now one should press "a" to generate the animation, instead of "q", which quit the rendering window to create a new one, which did not work on Linux.


ansys/dpf/core/animator.py
```diff
@@ -80,39 +81,51 @@ def render_field(frame):
...
+                                print("Animation canceled.")
+                                return result
+                        if save_as:
+                            self._plotter.write_frame()
+
+            # Write initial frame
+            render_frame(0)
+            # If not off_screen, enable the user to choose the camera position
+            off_screen = kwargs.pop("off_screen", None)
+            if off_screen is None:
```
anslpa, line 116: @PProfizi that means that we set off_screen by defualt. Is that what we want ?

PProfizi: @anslpa I do not see how this sets `off_screen=True` by default. This takes the value in `kwargs`, if not present then it is set to `None`, and if `None `then we do a last check of the `pv.OFF_SCREEN` variable which might have been set globally to `True`, for example when running tests or examples in our pipelines.
I tested it and to me it is `False `by default.

cbellot000 (approved):

PR #592, opened 2022-10-27
Fix GrpcServer.ip return type upon error

PProfizi:
Should return an empty string upon exception, and not 0.


PProfizi:
Solved in the meantime by #598

PR #606, opened 2022-11-08
Feature/expose context

cbellot000:


ansys/dpf/core/server_context.py
```diff
@@ -0,0 +1,101 @@
...
+    custom_defined = 3
+    """Loads the xml named "DpfCustomDefined.xml" that the user can modify."""
+    entry = 4
+    """Loads the minimum number of plugins for a basic usage."""
+
+
+class ServerContext:
+    """The context allows to choose which capabilities are available server side.
+
+    The default parameters can be overwritten using the DPF_SERVER_TYPE environment
```
PProfizi, line 28: @cbellot000 It is not clear to me from this sentence the link between class ServerContext and the server types.

ansys/dpf/core/server_context.py
```diff
@@ -0,0 +1,101 @@
...
+    DPF_SERVER_TYPE=LEGACYGRPC can be used.
+
+    Parameters
+    ----------
+    context_type : EContextType
+        Type of context.
+    xml_path : str, optional
+        Path to the xml to load.
+    """
+    def __init__(self, context_type=EContextType.user_defined, xml_path=""):
```
PProfizi, line 39: @cbellot000 Why is the user_defined context the default one? I think this makes the xml_path non-optional in a way.

ansys/dpf/core/server_context.py
```diff
@@ -0,0 +1,101 @@
...
+    """Loads the xml named "DpfCustomDefined.xml" that the user can modify."""
+    entry = ServerContext(EContextType.entry)
+    """Loads the minimum number of plugins for a basic usage. Is the default."""
+
+
+SERVER_CONTEXT = AvailableServerContexts.entry
+
+
+def apply_server_context(context=AvailableServerContexts.entry, server=None) -> None:
+    """Allows to apply a complex globally (if no server is specified) or to a
```
PProfizi, line 74:
```suggestion
    """Allows to apply a context globally (if no server is specified) or to a
```

ansys/dpf/core/server_context.py
```diff
@@ -0,0 +1,101 @@
...
+    given server.
+    When called before any server is started, the context will be applied to the next servers.
```
PProfizi, line 76:
```suggestion
    When called before any server is started, the context will be applied by default to any new server.
```

ansys/dpf/core/server_context.py
```diff
@@ -0,0 +1,101 @@
...
+
+    The context allows to choose which capabilities are available server side.
+
+    Parameters
+    ----------
+    context : ServerContext
+        Context to apply to the given server or to the next started servers (when no server
```
PProfizi, line 83:
```suggestion
        Context to apply to the given server or to the newly started servers (when no server
```

ansys/dpf/core/server_types.py
```diff
@@ -774,8 +778,14 @@ def __init__(self,
                     f"DPF directory not found at {os.path.dirname(path)}"
                     f"Unable to locate the following file: {path}")
             raise e
-        data_processing_capi.DataProcessingCAPI.data_processing_initialize_with_context(1, None)
         self.set_as_global(as_global=as_global)
+        try:
+            self._base_service.apply_context(server_context.SERVER_CONTEXT)
+        except errors.DpfVersionNotSupported:
+            self._base_service.initialize_with_context(
+                server_context.AvailableServerContexts.premium
```
PProfizi, line 786: @cbellot000 Why premium in this case?

tests/conftest.py
```diff
@@ -421,3 +440,11 @@ def count_servers():
         # assert num_dpf_exe == 1
 
     request.addfinalizer(count_servers)
+
+
+# to call at the end
+core.server.shutdown_all_session_servers()
+try:
+    core.apply_server_context(core.AvailableServerContexts.premium)
+except core.errors.DPFServerException:
```
PProfizi, line 449: @cbellot000 Can we not be a little bit more specific? To not catch any possible Server exception

tests/test_multi_server.py
```diff
@@ -8,6 +8,9 @@
 from ansys.dpf.core.server_factory import ServerConfig, CommunicationProtocols
 
 
+dpf.apply_server_context(dpf.AvailableServerContexts.entry)
```
PProfizi, line 11: @cbellot000 Is this to test applying the entry context? Or for the "other_remote_server" fixture to return "entry" GrpcServers? Could that not be directly set in the fixture?

tests/test_service.py
```diff
@@ -440,5 +440,61 @@ def test_load_api_with_awp_root_2_no_gatebin():
     assert dpf_inner_path in serv._grpc_client_path
 
 
+@pytest.fixture(autouse=False, scope="function")
+def clean_up(request):
```
PProfizi, line 444: @cbellot000 Could we call this something more explicit such as "set context back to premium" or "entry_context"?

tests/test_service.py
```diff
@@ -440,5 +440,61 @@ def test_load_api_with_awp_root_2_no_gatebin():
...
+            pass
+
+    request.addfinalizer(revert)
+
+
+@pytest.mark.order(1)
+@conftest.raises_for_servers_version_under("4.0")
+def test_apply_context(clean_up):
+    # Carefully: this test only work if the premium context has never been applied before on the
+    # in process server.
```
PProfizi, line 468:
```suggestion
    # in process server, otherwise premium operators will already be loaded.
```

ansys/dpf/core/server_context.py
cbellot000: @PProfizi it's the default of this constructor (because you would only need to use the constructor directly when you want to speicify an xml), it doesn't make user defined the default context (which is SERVER_CONTEXT = AvailableServerContexts.entry)

ansys/dpf/core/server_context.py
cbellot000: sorry, bad copy paste ;)

ansys/dpf/core/server_types.py
cbellot000: @PProfizi premium was the old default (equivalent to standalone)

tests/conftest.py
cbellot000: sorry it's a mistake

anslpa (approved):

PR #607, opened 2022-11-08
Update README.md install command for development

PProfizi:


PProfizi:
> I have seen in other repos that the README is included as the first page of the documentation. Maybe we could also do this here? So we avoid to have "duplicated" documentation?

Right now the two do not have the same structure. We might do this in the future but this is not a priority right now.

PR #635, opened 2022-11-22
Ask for Ansys version in bug issue template

PProfizi:
One critical information which was not in the bug issue template is the Ansys installation version being used to run DPF.


GuillemBarroso (approved):
LGTM

PR #644, opened 2022-11-24
Update the operators for 2023.2.pre0

PProfizi:
Use of .ci/code_generation.py to generate the python code for operators contained in DPF Server 2023.2.pre0

- [x] Check the server used to generate the operators is the latest release candidate
- [x] Check the list of plugins included in the list to generate operators for in code_generation.py
- [x] Check the list of operators, especially those which might have been suppressed


PProfizi:
@cbellot000 @anslpa @rafacanton the issue here seems to be that (as discussed briefly with Rafael) the init methods of the operators have been reworked and do not include all the possible inputs as was the case before.
Thus, asking for `dpf.operators.result.displacement(mesh=mesh)` does for example not work, just like `dpf.operators.result.stress_von_mises(requested_location="Nodal")` and many others.
What is the new strategy here? What is the new best practice? Do we go back to using `op.connect()` for any pin not in the default operator constructor?  
For reference, all constructors (at least of the result section) of the operators now only have these arguments:
`__init__(self, time_scoping=None, mesh_scoping=None, data_sources=None, config=None, server=None)`

rafacanton:
@PProfizi, @cbellot000, @anslpa I guess that we don't have any other option than connecting thorugh the `connect` method. Personally, I don't dislike it. It makes more explicit the concept of operator with input pins and how can Dpf be used to build workflows

.ci/code_generation.py
```diff
@@ -58,17 +61,22 @@
             os.remove(f)
     except:
         pass
-core.start_local_server(config=core.AvailableServerConfigs.LegacyGrpcServer)
+core.start_local_server(config=core.AvailableServerConfigs.GrpcServer)
 code_gen = core.Operator("python_generator")
```
rafacanton, line 65: This will only work for PREMIUM. Should we explicitly add it as context?

PProfizi: @rafacanton, sure, we can add it explicitely at the beginning of the script.
Also, sorry about the confusion Rafael, the problems with the constructors of the operators is that there is a second step in the operator generation, which I now added to the script.

anslpa (approved):

PR #663, opened 2022-12-01
Adapt to pre-commit.yaml file

Revathyvenugopal162:
Add pre-commit.yaml file for the code style check.


GuillemBarroso:
Thanks for working on this @Revathyvenugopal162. 

The build is failing due to a circular error:
image

This happened to me before in the other PR I had open. I think it's related to the ordering of the imports in ``__init__.py``. Is this possible?

If that is the case, we need to find the way to have the imports sorted alphabetically without breaking everything.

Revathyvenugopal162:
> Thanks for working on this @Revathyvenugopal162.
> 
> The build is failing due to a circular import error: image
> 
> This happened to me before in the other PR I had open. I think it's related to the ordering of the imports in `__init__.py`. Is this possible?
> 
> If that is the case, we need to find the way to have the imports sorted alphabetically without breaking everything.

Yeah, we have isort, we should add that for the sorting of import. let me see.

PR #670, opened 2022-12-05
Context: status now handled on server side

anslpa:


PProfizi:
Hi @anslpa, can we close this as stale?

PR #673, opened 2022-12-06
DPF Server: default server config issue 

anslpa:
If ansys_dpf_server_win_v2023.2.pre0 is used and this is contained in the server folder name, v202 is parsed as Ansys version and GrpcLegacyServer is the default. This prevents this behavior to have InProcessServer instead.


cbellot000 (approved):

PProfizi (approved):

anslpa:
It was needed to add check to ensure ansys_path was not None, because None type is not iterable and that's what is gotten for run on Docker (CI/CD build issue).

PR #691, opened 2022-12-14
Add geometry objects for plotting

### LINKED ISSUE #690: Geometric objects for plotting

Define some standard geometries so they can be used to map field values for plotting purposes.

The first basic geometries would be:
- Points
- Lines
- Planes

Further operations (union, instersection) are planned to be included in the future. 

Create classes for each geometric entity. 

The case of lines and planes will require from a discretization in order to define 1D (line) and 2D (plane) meshes.

Provide example on how to use it.

_No response_

GuillemBarroso:
Fix #690.

This PR proposes a few basic geometric objects (points, lines and planes) to ease the user interaction when plotting sections, line plots, etc.

A test suite for the new implementation is provided.

Also, a plotting example has been added to demonstrate how to use these geometrical entities when plotting.


GuillemBarroso:
@PProfizi, @cbellot000, @germa89, @rlagha, feedback will be greatly appreciated!

I still need to add some documentation and examples on the new classes, but I think this PR shouldn't include much more new stuff.

PProfizi (approved):
LGTM! TBC!

src/ansys/dpf/core/geometry.py
```diff
@@ -0,0 +1,403 @@
...
+    @property
+    def dimension(self):
+        """Dimension of the Points object space."""
+        return 3
+
+    def plot(self, **kwargs):
+        """Visualize Points object."""
+        pl = DpfPlotter(**kwargs)
+        pl.add_points(self._coordinates.data)
+        pl.show_figure(show_axes=True)
```
germa89, line 88: Does this accept to input our own plotter?? Should we return the plotter somehow?

src/ansys/dpf/core/geometry.py
```diff
@@ -0,0 +1,403 @@
...
+
+    >>> from ansys.dpf.core.geometry import Line
+    >>> line = Line([[0, 0, 0], [1, 0, 0]])
+    >>> print(line)
+    DPF Line object:
+    Starting point: [0. 0. 0.]
+    Ending point: [1. 0. 0.]
+    Line discretized with 100 points
+    >>> line.plot()
+
```
germa89, line 114: Good docstrings!

src/ansys/dpf/core/geometry.py
```diff
@@ -0,0 +1,403 @@
...
+    """
+
+    def __init__(self, coordinates, num_points=100, server=None):
+        """Initialize line object from two 3D points and discretize."""
+        if not isinstance(coordinates, Field):
+            coordinates = field_from_array(coordinates)
+        if not len(coordinates.data) == 2:
+            raise ValueError("Only two points have to be introduced to define a line")
```
germa89, line 122:
```suggestion
            raise ValueError("Only two points must be introduced to define a line")
```

src/ansys/dpf/core/geometry.py
```diff
@@ -0,0 +1,403 @@
...
+    >>> print(line)
+    DPF Line object:
+    Starting point: [0. 0. 0.]
+    Ending point: [1. 0. 0.]
+    Line discretized with 100 points
+    >>> line.plot()
+
+    """
+
+    def __init__(self, coordinates, num_points=100, server=None):
```
germa89, line 117: I dont like to have ``num_points`` hardcoded.... Also I dont see why we need to have the line discretized. I think it is for the intersection with other objects, but I would expect pyvista would take care of that.

src/ansys/dpf/core/geometry.py
```diff
@@ -0,0 +1,403 @@
...
+        ]
+        return diff / np.linalg.norm(diff)
+
+    def plot(self, **kwargs):
+        """Visualize line."""
+        # Check if line is along the [1, 1, 1] direction to change camera position
+        if np.isclose(
+            np.abs(self.direction), normalize_vector(np.array([1, 1, 1]))
+        ).all():
+            camera_position = "xy"
```
germa89, line 205: Good point!

src/ansys/dpf/core/geometry.py
```diff
@@ -0,0 +1,403 @@
...
+        else:
+            camera_position = None
+
+        # Plot line object
+        pl = DpfPlotter(**kwargs)
+        pl.add_line(self._coordinates.data)
+        pl.show_figure(show_axes=True, cpos=camera_position)
```
germa89, line 212: Same concerns about ``points.plot``

src/ansys/dpf/core/geometry.py
```diff
@@ -0,0 +1,403 @@
...
+
+
+class Plane:
+    """
```
germa89, line 216: Nice doc strings!

src/ansys/dpf/core/geometry.py
```diff
@@ -0,0 +1,403 @@
...
+            txt += f"  Num cells y-dir: {self.num_cells_y}\n"
+        else:
+            txt += "Plane has not been discretized.\n"
+            txt += "  Use plane.discretize(width, height, num_cells_x, num_cells_y)\n"
+        return txt
+
+    @property
+    def center(self):
+        """Center of the plane."""
+        return self._center
```
germa89, line 299: A plane has center? I thought they were infinite.

src/ansys/dpf/core/geometry_factory.py
```diff
@@ -0,0 +1,133 @@
...
+
+def create_plane_from_center_and_normal(center, normal, server=None):
+    return Plane(center, normal, server)
+
+
+def create_plane_from_points(points, server=None):
+    """Create plane from three points."""
+    # Input check
+    if isinstance(points, Points):
+        if not len(points) == 3:
```
germa89, line 53: Makes me thing if we should have a method like ``point.is_in_plane(plane)`` or better ... ``point in plane``

src/ansys/dpf/core/plotter.py
```diff
@@ -87,6 +87,39 @@ def add_scale_factor_legend(self, scale_factor, **kwargs):
...
+        import pyvista as pv
+
+        plane = pv.Plane(
+            center=center,
+            direction=direction,
+            i_size=0.005,
+            j_size=0.005,
+            i_resolution=20,
+            j_resolution=20,
+        )
```
germa89, line 118: Shouldn't we use here the discretization?

germa89 (approved):
LGTM.

src/ansys/dpf/core/geometry.py
GuillemBarroso: I saw the ``.plot()`` method of each oject as a quick way for the user to check that this is indeed the object they intended to creats. What would you like to do when returning the plotter?

src/ansys/dpf/core/geometry.py
GuillemBarroso: Yes, but I decided to define a plane given its center and normal. Hence, we need to store and retriever these two variables from the object.

src/ansys/dpf/core/geometry_factory.py
GuillemBarroso: That could be useful, but I haven't seen where to use it yet.

src/ansys/dpf/core/plotter.py
GuillemBarroso: The discretized objects are just to interpolate there later. Conceptually, I think a user should be able to see t'he plane even if It has nit been discretized.

Having said that, I was planning on changing this and discretize all objects right after discretizing. The reason being is that, for the planes, it makes a lot of sense asking to users to provide width and height of the plane when they try to create it using a center and a normal. However, if they define 3 points I think It would be more intuitive to consider these 3 points as the corners of the plane, thus, width and height being already defined.

What do you think?? Of course I will allow users to change the width and height of the plane anyway.

If I go with this approach, I Will be able to always plot the plane with its defined width and height and I will modify these lines.

cbellot000 (approved):
Very useful new features Guillem, thank you!

src/ansys/dpf/core/geometry.py
germa89: I agree with doing the intersection before.

If it is for mapping, I'm fine with that.

src/ansys/dpf/core/geometry.py
germa89: A very valid reason!

src/ansys/dpf/core/plotter.py
germa89: > discretize all objects right after discretizing 

I'm not sure I follow.

> asking to users to provide width and height of the plane when they try to create it using a center and a normal. However, if they define 3 points I think It would be more intuitive to consider these 3 points as the corners of the plane, thus, width and height being already defined.

This is not very clear to me. What do you mean with `if they define 3 points I think It would be more intuitive to consider these 3 points as the corners of the plane, thus, width and height being already defined.` ?

GuillemBarroso: What I have implemented now is that the objects are discretized right after initialization, so the user will not need to first create a plane and then discretize it. It will be done all in one line.

Regarding the plane. When creating a plane with its center and normal, it makes sense to ask the user to provide width and height of the line as:
```py
plane = create_plane_from_center_and_normal(center=[0,0,0], normal=[0,1,0], width=1, height=1)
```

However, I think that when defining a plane with three points:
```py
plane = create_plane_from_points([[0,0,0], [0,2,0], [0, 0, 4]])
```
the user may expect these three points to be the corner points of the plane. Thus, the width and height will be computed accordingly. For instance, the previous case will return a plane that has width = 2 and height = 4.

PR #695, opened 2022-12-16
UnitSystems supported in Python

rafacanton:
With this PR, Dpf Unit Systems are now supported in Python through the `UnitSystem` class in `unit_system.py`. A set of predefined Ansys Unit Systems (the ones found in solution files from mechanical) are provided in the `unit_systems` Enum.

In addition, LSDyna source operators are updated to support custom unit systems.


PProfizi (approved):

rafacanton:
Summarizing what is done in this PR:

- Support is added for custom Unit Systems (`UnitSystem` class), and the `unit_systems` class is provided for the Ansys predefined unit systems.
- Tests are added to verify the custom Unit Systems.
- LS-Dyna tests and examples are improved to capture the Unit Systems behavior.
- The `unit_system` pin in the LSDyna operators is coherently overloaded for the three use cass (`int`, `str` or `UnitSystem`).

PR #707, opened 2022-12-19
Migrate to pyproject.toml files

Revathyvenugopal162:


pyproject.toml
```diff
@@ -0,0 +1,54 @@
+[build-system]
+requires = ["flit_core >=3.2,<4"]
+build-backend = "flit_core.buildapi"
+
+[project]
+# Check https://flit.readthedocs.io/en/latest/pyproject_toml.html for all available sections
+name = "ansys-dpf-core"
+version = "0.7.dev0"
```
GuillemBarroso, line 8: @PProfizi, how should we define ``version`` here? This is following other pyansys repos convention, but I remember we talked about PyDPF-Core using 0.7.Z.dev0, right?

pyproject.toml
```diff
@@ -0,0 +1,54 @@
...
+
+[project.optional-dependencies]
+plotting = ["pyvista>=0.32.0", "matplotlib>=3.2"]
+
+[tool.flit.module]
+name = "ansys.dpf.core"
+
+[project.urls]
+Homepage = "https://dpfdocs.pyansys.com/"
+Documentation = "https://dpfdocs.pyansys.com/"
```
GuillemBarroso, line 43:
```suggestion
Documentation = "https://dpf.docs.pyansys.com/"
```

pyproject.toml
```diff
@@ -0,0 +1,54 @@
...
+]
+
+[project.optional-dependencies]
+plotting = ["pyvista>=0.32.0", "matplotlib>=3.2"]
+
+[tool.flit.module]
+name = "ansys.dpf.core"
+
+[project.urls]
+Homepage = "https://dpfdocs.pyansys.com/"
```
GuillemBarroso, line 42:
```suggestion
Homepage = "https://dpf.docs.pyansys.com/"
```

GuillemBarroso (changes requested):
Some minor changes. LGTM! Let's wait for @PProfizi review as well.

pyproject.toml
PProfizi: @GuillemBarroso @Revathyvenugopal162 Indeed, the version should be 0.7.3.dev0 (we are releasing 0.7.2)

PProfizi, line 8:
```suggestion
version = "0.7.3.dev0"
```

pyproject.toml
```diff
@@ -0,0 +1,54 @@
...
+requires = ["flit_core >=3.2,<4"]
+build-backend = "flit_core.buildapi"
+
+[project]
+# Check https://flit.readthedocs.io/en/latest/pyproject_toml.html for all available sections
+name = "ansys-dpf-core"
+version = "0.7.dev0"
+description = "Data Processing Framework - Python Core "
+readme = "README.md"
+requires-python = ">=3.7"
```
PProfizi, line 11:
```suggestion
requires-python = ">=3.7, <3.11"
```
Not sure this is required but it might be?

PProfizi (commented):
I am alright with what I saw in the pyproject.toml
Do you know why black went and edited all files again? A change with the new version?
Also we might want to publish on testPyPI to make sure the metadata is alright.

pyproject.toml
Revathyvenugopal162: Ok i got your point. but usually we will not keep track of the patch releases, 0.7.dev0 or if we have already 0.7 release, then 0.8.dev0 is the usual way of keeping track, so that the patch release is not going to affect the dev version. https://peps.python.org/pep-0440/#developmental-releases. but no offence. ultimately the developers can decide .

Revathyvenugopal162:
> I am alright with what I saw in the pyproject.toml Do you know why black went and edited all files again? A change with the new version? Also we might want to publish on testPyPI to make sure the metadata is alright.

The default line length formatting for black is 80. but in flake 8 we specified to 100, so i changed to 100. And i think it is a good number.

pyproject.toml
GuillemBarroso: I'm fine with either. I think it would be nice to be consistent with other repos and use 0.8.dev0, but I am not fully aware of the implications of changing this. @PProfizi, your call.

GuillemBarroso:
> > I am alright with what I saw in the pyproject.toml Do you know why black went and edited all files again? A change with the new version? Also we might want to publish on testPyPI to make sure the metadata is alright.
> 
> The default line length formatting for black is 80. but in flake 8 we specified to 100, so i changed to 100. And i think it is a good number.

Yep. I like 100.

PProfizi (approved):

PProfizi (approved):

PR #708, opened 2022-12-19
FIX: doc-deploy-dev trigger

jorgepiloto:
It looks like the development documentation has not been triggering, see https://github.com/pyansys/pydpf-core/actions/runs/3731763945.

Changing here the conditional. Do you have any clues on why this may be happening, @PProfizi?


PProfizi:
@jorgepiloto actually no I really do not know why it would not trigger...
We are also having issues with whole pipelines not triggering...

PProfizi (approved):

PR #714, opened 2022-12-20
Fix geometry tests

GuillemBarroso:
There was an import issue detected by @rafacanton. 

I have removed the variables I used for testing and place them inside the decorators ``@pytest.mark.parametrize``. This should prevent creating the lists when importing the file.

### DIFF
--- a/tests/test_geometry.py
+++ b/tests/test_geometry.py
@@ -32,22 +32,22 @@ def test_create_points():
     assert len(points) == points.n_points == n_points
 
 
-points_data = [
-    ([[0.4, 0.1, 0], [0.1, 0, 0.5]]),
-    ([[0.0, 0.0, 0.0], [1.0, 1.0, 1.0]]),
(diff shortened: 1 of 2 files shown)


rafacanton:
@cbellot000, @PProfizi This one refactors a bit the geometry and multi_server tests. It fixes part of the issue we had on the standalone and Lin Daily build. There is a second issue remaining that relates to the fact that we are adding explicit `.plot()` calls in all geometry tests. This means that these tests cannot be run in TFS unless we do a workaround...

@GuillemBarroso proposed [this]([actions/action.yml at main · pyansys/actions (github.com)](https://github.com/pyansys/actions/blob/main/tests-pytest/action.yml#L56)) workaround

PProfizi:
Hey @rafacanton, I am not sure why the plot commands would be a problem though. We have plot commands in other tests without problems. What would be specific here?

GuillemBarroso:
Closing this PR as #716 solved this issue already.

PR #716, opened 2022-12-21
fix test_geometry

cbellot000:


GuillemBarroso (approved):
Great @cbellot000! Thanks for working on that. 

Just curious about the lambda. Is that how you achieve that Points, Line and Plane are only evaluated within the test and not at the import?

cbellot000:
> evaluated

Yes exactly, it's ran in the test itself

PR #724, opened 2022-12-22
Intersection of geometry objects with meshed_regions

### LINKED ISSUE #692: Get intersections with mesh when using geometric objects

The Plotting on geometry elements example implemented in this [PR](https://github.com/pyansys/pydpf-core/pull/691) plots the result of the mapping operator ``on_coordinates``. 

However, in cases where a line or a plane is both inside and outside of the geometry, only the inner points can be mapped. This leads to plots like the following: 

image

As an improvement, the intersection points should also be plotted so the plot also displays the interface plane/mesh.

1. Find points in the intersection.
2. Use the mapping operator ``prepare mapping workflow`` to get the solution in the intersection points.
3. Add intersection points and its associated field values to the plane
4. Plot plane. 

N/A

GuillemBarroso:
Fix #692.

Add examples on how to intersect points, lines and planes objects with a 3D mesh.


examples/05-plotting/07-plot_on_points.py
```diff
@@ -0,0 +1,88 @@
...
+# on_coordinates>`: mapping opretor
+disp = model.results.displacement
+mapping_operator = ops.mapping.on_coordinates(
+    fields_container=disp,
+    coordinates=field_from_array(points.coordinates.data),
+    create_support=True,
+    mesh=mesh,
+)
+fields_mapped = mapping_operator.outputs.fields_container()
+field_points = fields_mapped[0]
```
MaxJPRey, line 79: Maybe here we could explain/remember why we only select the first fields.
Just a reminder for people beginning  to use DPF.

GuillemBarroso: Good point @MaxJPRey ! Let me add a comment clarifying this. Thanks.

PR #726, opened 2022-12-23
Add example on how to get data by ID

### LINKED ISSUE #725: Add example on how to get data by ID
### Description of the modifications

Thanks to user's feedback, we have identified the need to add a new example.

This example should show how to efficiently retrieve data of a field by node ID after scoping on a specific named selection of the model. 

N/A

GuillemBarroso:
Fix #725.

This PR proposes a new example showing how to retrieve data from a field by ID. The example follows:

1. Load model and show the entire mesh.
2. Create nodal scoping in only one available named selection.
3. Apply scoping to displacements and mesh
4. Show mesh for that named selection.
5. Show how to access node ID and value for that node of the scoped displacement field.

### DIFF
--- a/examples/00-basic/13-get_data_by_id.py
+++ b/examples/00-basic/13-get_data_by_id.py
@@ -32,12 +32,14 @@ print(model.metadata.available_named_selections)
 ###############################################################################
 # Visualize the entire mesh
 # ~~~~~~~~~~~~~~~~~~~~~~~~~
-# Extract displacements on the entire geometry
-# .. note::
-#     Note that this is just for demonstration purposes. On a real simulation, the
-#     user shouldn't evaluate (request from the server) data that is not needed.
+# Extract displacements on the entire geometry. Note that this is just for
+# demonstration purposes. On a real simulation, the user should not evaluate
+# (request from the server) data that is not needed.
 disp_fc = model.results.displacement().eval()
 print(disp_fc)
+
+###############################################################################
+# The displacement FieldsContainer contains a total of 15113 entities.
 
 ###############################################################################
 # Plot the entire mesh
@@ -65,12 +67,15 @@ disp_selection_fc = model.results.displacement.on_mesh_scoping(mesh_scoping).eva
 print(disp_selection_fc)
 
 ###############################################################################
+# Note how the number of entities has decreased from 15113 to 12970 entities
+# after applying the scoping on ``my_named_selection``.
+
+###############################################################################
 # Get the only Field available
 disp_selection = disp_selection_fc[0]
 
 ###############################################################################
 # Scope mesh only for that named selection
-# mesh_scoping = model.metadata.named_selection(my_named_selection)  ## Force location == nodal
 mesh_from_scoping_op = ops.mesh.from_scoping()
 mesh_from_scoping_op.inputs.scoping.connect(mesh_scoping)
 mesh_from_scoping_op.inputs.mesh.connect(model.metadata.meshed_region)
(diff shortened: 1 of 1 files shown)


GuillemBarroso:
@rlagha, a couple comments:
- It would be great to verify that this is indeed the most efficient way to get data by node ID inside a loop. How this approach compares to the ``get_entity_data_by_id`` method? Do we have access to an example with a big mesh (0.5, 1, 2 million nodes)? Then I could assess both approaches in a more realistic setting.
- We have to think how we expose this in PyDPF-Post. I would implement the ``by_id`` method to be applied to a certain field. 
  Then, we could return two arrays with the ``nodal_ids`` and ``nodal_disp`` information as
  ```py
  simulation = post.load_simulation("my_example")
  my_named_selection = "test"
  my_selection = select(named_selection=my_named_selection)
  nodal_ids, nodal_disp = simulation.displacement(select=my_selection).by_id()
  assert len(nodal_ids) == len(nodal_disp) == n_nodes
  assert type(nodal_ids) == np.ndarray or ints
  assert type(nodal_disp) == np.ndarray of floats
  ```
  or return a single displacement value associated to a certain node ID as
  ```py
  simulation = post.load_simulation("my_example")
  my_named_selection = "test"
  my_selection = select(named_selection=my_named_selection)
  nodal_id, nodal_disp = simulation.displacement(select=my_selection).by_id(5)
  assert type(nodal_ids) == int
  assert type(nodal_disp) == np.array of floats # (containing the three components of the displacement for that node)
  ```
Thoughts?

examples/00-basic/13-get_data_by_id.py
```diff
@@ -0,0 +1,173 @@
...
+
+###############################################################################
+# Access data in ``disp_selection`` by ID (get entity approach)
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# An alternative to the index approach is to use the ``get_entity_data_by_id``
+# method to get the nodal displacement given a certain node ID.
+start = time.time()
+nodal_disp_get_entity = np.zeros([mesh_selection.nodes.n_nodes, 3])
+for idx, node_id in enumerate(ids):
+    node_disp = disp_selection.get_entity_data_by_id(node_id)
```
rlagha, line 153: @Guillem , this one is useless in this case. the right one in the one you wrote above. 
we use the get_by_id:
-when we loop over several fields for example nodal stresses and displacements and want to get for a given node of the displacement field it's nodal stress for example.
-when we have a user defined scoping, which is different from the field scoping, and we want to extract data from the field. this can be important when we want to read the data once from the file in one request and then work on different named selections and extract data from the field instead of reading each time from a file.

GuillemBarroso: Hi @rlagha. Yes, exactly. 

Depending on what you want to do you should use one approach or the other. That is why I included the two of them in the example and a comparison between them.

rlagha: @Guillem, this can be important to mention these remarks in the example.

examples/00-basic/13-get_data_by_id.py
```diff
@@ -0,0 +1,173 @@
...
+# before the loop below so only one call is made to the server.
+start = time.time()
+data = disp_selection.data
+
+###############################################################################
+# Loop over ``ids`` to have access to both, the node ID and the displacement
+# for that node
+nodal_disp_index = np.zeros([mesh_selection.nodes.n_nodes, 3])
+for idx, node_id in enumerate(ids):
+    node_disp = data[idx]
```
rlagha, line 140: @Guillem, this is true when you have nodal or elemental results. When you read stresses for example as Elemental nodal data, then you will need to use field.get_entity_data(idx) instead of data[idx], otherwise you will get only the first stress tensor, this is also true for layered results. 
We may need to add a similar loop with field.get_entity_data(); when the server is in in-process mode I think that you will not have overhead, if the server is remote then you may need to use as_local_field option and field.get_entity_data()

examples/00-basic/13-get_data_by_id.py
```diff
@@ -0,0 +1,173 @@
...
+# value indicates that the two approaches are completely equivalent.
+diff = np.sum(nodal_disp_index - nodal_disp_get_entity)
+print(f"Difference between the two approaches: {diff}")
+
+###############################################################################
+# In terms of performance, the index approach is found to be an order of magnitude
+# faster for this particular example. The time difference can be attributed to the
+# fact that only one server call is required for the index approach.
+# Instead, the get entity strategy requires one server call per node in the mesh.
+# The difference in perforance is likely to scale with the number of nodes in the mesh.
```
MaxJPRey, line 171:
```suggestion
# The difference in performance is likely to scale with the number of nodes in the mesh.
```

PR #750, opened 2023-01-10
Add more info on the fields'data ordering

cbellot000:


src/ansys/dpf/core/field.py
```diff
@@ -98,6 +106,31 @@ class Field(_FieldBase):
     >>> field.data[2]
     DPFArray([-0.00672665, -0.03213735,  0.00016716]...
 
+    Accessing data with a custom order.
+
+    >>> from ansys.dpf import core as dpf
+    >>> from ansys.dpf.core import examples
+    >>> transient = examples.download_transient_result()
+    >>> model = dpf.Model(transient)
+    >>> ids_order = [2,3]
```
PProfizi, line 115: @cbellot000  isn't it `ids_order = [3, 2]` in that case?

MaxJPRey:
LGTM.
Just a question not related directly to this PR for my own knowledge.
The docstring test is failing for Linux but not on Windows.
image

Why do we get the `dtype` -int32 in our case- in `DPFArray` when calling field.scoping.ids in Linux and not Windows?

src/ansys/dpf/core/field.py
cbellot000: no @PProfizi that's exactly what I want to show case here: you ask [2,3] and you might end up with [3,2]

PProfizi (approved):

PR #758, opened 2023-01-18
Fix and improve plotter.DpfPlotter behavior

PProfizi:
Solves #767 

DpfPlotter.add_node_labels now correctly draws the bare meshed_region given in input, with labels, in the case where no Field was added to the plotter first.
Improved typehinting and docstrings
DpfPlotter.show_figure now warns the user that the `notebook` kwarg is to be given at initialization of the DpfPlotter.

### DIFF
--- a/src/ansys/dpf/core/plotter.py
+++ b/src/ansys/dpf/core/plotter.py
@@ -21,9 +21,9 @@ from ansys.dpf import core
 from ansys.dpf.core.common import locations, DefinitionLabels
 from ansys.dpf.core.common import shell_layers as eshell_layers
 from ansys.dpf.core import errors as dpf_errors
+from ansys.dpf.core.nodes import Node, Nodes
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


src/ansys/dpf/core/plotter.py
```diff
@@ -149,9 +156,20 @@ def add_mesh(self, meshed_region, deform_by=None, scale_factor=1.0, **kwargs):
...
+    def add_point_labels(
+        self,
+        nodes: Union[Nodes, List[Node], List[int]],
+        meshed_region: MeshedRegion,
+        labels: Union[List[str], None] = None,
+        **kwargs,
+    ) -> List:
         label_actors = []
-        node_indexes = [meshed_region.nodes.mapping_id_to_index.get(node.id) for node in nodes]
+        if isinstance(nodes, Nodes):
```
rafacanton, line 167: You could maybe add another test with `nodes` being a `Nodes` and silence this Codecov warning

rafacanton (approved):

PR #773, opened 2023-01-30
Allow for Python <4 versions

RobPasMue:
When moving from ``pyproject.toml`` to ``setup.py``, PyDPF-Core stopped supporting Python 3.11+ versions. In your previous ``setup.py`` file, your Python limitations were ``>=3.7, <4`` and now they are ``>=3.7, <3.11``. This will cause a conflict with the PyAnsys metapackage as we can tell from: https://github.com/pyansys/pyansys-dev/issues/4

I submit this PR to replicate the previous behavior you had. I understand that this change was unintended. Consider adding a Python 3.11 smoke test in your workflow as well =)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ name = "ansys-dpf-core"
 version = "0.7.3.dev0"
 description = "Data Processing Framework - Python Core "
 readme = "README.md"
-requires-python = ">=3.7, <3.11"
+requires-python = ">=3.7, <4"
 license = {file = "LICENSE"}
 authors = [
     {name = "ANSYS, Inc.", email = "ramdane.lagha@ansys.com"},


Revathyvenugopal162 (approved):
Thanks for solving this @RobPasMue .

PProfizi:
Hey @RobPasMue, thanks for this change, although I do not think it was a mistake. We initially did not support 3.11 explicitly, hence the lack of testing. What I understand though is that pyansys packages should explicitly support 3.11 or at the very least allow it.
What I will do is merge this PR so we do not prevent the pyansys metapackage CI from passing, yet I will also create an issue here to add explicit testing of 3.11. I am not fond of just doing smoke tests, so we usually test all Python versions we are supposed to support. We should most likely move to running only examples though.

RobPasMue:
> Hey @RobPasMue, thanks for this change, although I do not think it was a mistake. We initially did not support 3.11 explicitly, hence the lack of testing. What I understand though is that pyansys packages should explicitly support 3.11 or at the very least allow it. What I will do is merge this PR so we do not prevent the pyansys metapackage CI from passing, yet I will also create an issue here to add explicit testing of 3.11. I am not fond of just doing smoke tests, so we usually test all Python versions we are supposed to support. We should most likely move to running only examples though.

Oh great then. Yeah, we are avoiding explicit disregard of Python 3.11 - at least allow it even if you are not sure if it works perfectly fine or not. If you intend on testing all the Python versions you support that's even better :)

PR #783, opened 2023-02-06
Add minimum DPF version logic to examples

PProfizi:
Managing the minimal DPF version required to run an example is necessary to be able to continue CI/CD with continuous integration of examples and features only available on the DPF server currently under development, but not yet in the one already out.
For now, this is specified with a note in the header of the example, stating for example:
 ```
.. note::
    This example requires DPF 6.1 or above.
    For more information, see :ref:`ref_compatibility`.
```
The two digits version number is used, however I think we could use the DPF Server package version if we wanted to.

See [docs](https://github.com/pyansys/pydpf-core/actions/runs/4105649624) job as well as [examples](https://github.com/pyansys/pydpf-core/actions/runs/4105653287) job for the 2023.2.pre0 server.

### DIFF
--- a/.ci/run_examples.py
+++ b/.ci/run_examples.py
@@ -4,12 +4,19 @@ import pathlib
 import subprocess
 import sys
 
+import ansys.dpf.core as dpf
+
 os.environ["PYVISTA_OFF_SCREEN"] = "true"
 os.environ["MPLBACKEND"] = "Agg"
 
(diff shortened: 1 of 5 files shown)


PProfizi:
Tested locally and works for gallery generation, like here with 00-basic_example.py declared as requiring 6.2 in a note:
image

cbellot000 (approved):

PR #797, opened 2023-02-13
Allow doctest to deal with features requiring future server version

PProfizi:
Features added to PyDPF-Core only available with the DPF server currently under development should not prevent the CI_release from running and PyDPF-Core from being able to be released anytime.
Those are marked as requiring DPF server version > XXX.
Tests are versions, examples now also deal with that, yet the docstrings examples (run through doctest) were not managed.

This PR adds a [custom doctest runner](https://docs.python.org/3/library/doctest.html#advanced-api) to detect docstring examples failing due `DpfVersionNotSupported`, and mark them as passing.
Before: https://github.com/pyansys/pydpf-core/actions/runs/4156462538/jobs/7190270043#step:8:89
After: https://github.com/pyansys/pydpf-core/actions/runs/4175995794

### DIFF
--- a/src/ansys/dpf/core/unit_system.py
+++ b/src/ansys/dpf/core/unit_system.py
@@ -118,11 +118,14 @@ class unit_systems:
 
     """
 
-    solver_mks = UnitSystem("solver_mks", ID=11)
-    solver_cgs = UnitSystem("solver_cgs", ID=5)
-    solver_nmm = UnitSystem("solver_nmm", ID=6)
-    solver_umks = UnitSystem("solver_umks", ID=10)
-    solver_knms = UnitSystem("solver_knms", ID=16)
-    solver_bft = UnitSystem("solver_bft", ID=7)
-    solver_bin = UnitSystem("solver_bin", ID=8)
-    undefined = UnitSystem("undefined", ID=-1)
+    try:
+        solver_mks = UnitSystem("solver_mks", ID=11)
+        solver_cgs = UnitSystem("solver_cgs", ID=5)
+        solver_nmm = UnitSystem("solver_nmm", ID=6)
+        solver_umks = UnitSystem("solver_umks", ID=10)
+        solver_knms = UnitSystem("solver_knms", ID=16)
+        solver_bft = UnitSystem("solver_bft", ID=7)
+        solver_bin = UnitSystem("solver_bin", ID=8)
+        undefined = UnitSystem("undefined", ID=-1)
+    except dpf_errors.DpfVersionNotSupported as e:
+        pass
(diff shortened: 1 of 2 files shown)


PProfizi:
See https://github.com/pyansys/pydpf-core/actions/runs/4173892954 for working Test jobs with standalone .pre0 (UnitSystem not yet supported)

src/ansys/dpf/core/unit_system.py
```diff
@@ -95,34 +95,39 @@ def unit_names(self) -> str:
             return self._unit_names
 
 
-class unit_systems:
-    """Contains common Ansys predefined UnitSystems.
+try:
```
rafacanton, line 98: @PProfizi This means that starting from now we will need to add `try/catch` blocks in all code that we add? Not talking about tests, just the source code. I think this will decrease the readability...

PProfizi: @rafacanton indeed, I required this locally when running doctest specifically on this file. I guess doctest first tried to execute the file, which failed for 6.0. 
I'll try without.

PProfizi: @cbellot000 @anslpa Just to let you know, after discussing this with Rafael, we decided that the right pattern is to put a "Required Version" note for the "helper" `unit_systems` class just like for the "feature" `UnitSystem` class, as well as a try-catch when populating it, so that the `unit_systems` class may appear for previous server versions, but it will be empty and the doc will indicate why.
Do you think this pattern is ok?

rafacanton: @PProfizi, @cbellot000, @anslpa we could maybe stablish this kind of pattern for the future when this happens again

rafacanton (approved):

src/ansys/dpf/core/unit_system.py
```diff
@@ -118,11 +122,14 @@ class unit_systems:
...
+    try:
+        solver_mks = UnitSystem("solver_mks", ID=11)
+        solver_cgs = UnitSystem("solver_cgs", ID=5)
+        solver_nmm = UnitSystem("solver_nmm", ID=6)
+        solver_umks = UnitSystem("solver_umks", ID=10)
+        solver_knms = UnitSystem("solver_knms", ID=16)
+        solver_bft = UnitSystem("solver_bft", ID=7)
+        solver_bin = UnitSystem("solver_bin", ID=8)
+        undefined = UnitSystem("undefined", ID=-1)
+    except dpf_errors.DpfVersionNotSupported as e:
```
PProfizi, line 134:
```suggestion
    except dpf_errors.DpfVersionNotSupported as e:  # pragma: no cover
```

cbellot000 (approved):

PR #810, opened 2023-02-23
Fix Line.__init__ for Python 3.7 with numpy 1.21.6

PProfizi:
Translate coordinates to numpy array of type np.number in Line.init when not a DPF Field.
Fixes an error here https://github.com/pyansys/pydpf-core/actions/runs/4252250685/jobs/7395598740#step:11:2539
specific to Python 3.7 which still uses Numpy 1.21.6 (latest for 3.7).


PProfizi:
Test CI equivalent to a release context: https://github.com/pyansys/pydpf-core/actions/runs/4252968621

anslpa (approved):

cbellot000 (approved):

PR #834, opened 2023-03-10
Update email contacts in package metadata

PProfizi:


jorgepiloto (approved):
Thanks for updating this, @PProfizi. Approving 🚀

PR #846, opened 2023-03-17
fix unit tests for 2023.2.pre1 server and add License context manager

cbellot000:


src/ansys/dpf/core/server_context.py
```diff
@@ -38,6 +40,111 @@ def same_licensing_context(first, second):
         return True
 
 
+class LicenseContextManager:
+    """Can be optionally used to pre checkout a licence before using licensed DPF Operators.
+    Improves performances if many Operators requiring licensing are used afterwards.
+    It can also be used to force checkout before running a script when few
+    ANSYS license increments are available.
+    Check in the license when the object is deleted.
```
JennaPaikowsky, line 48:
```suggestion
    """Can optionally be used to check out a license before using licensed DPF Operators.
    Improves performance if you are using multiple Operators that require licensing.
    It can also be used to force checkout before running a script when few
    Ansys license increments are available.
    The license is checked in when the the object is deleted.
```

JennaPaikowsky, line 48: For line 48, my assumption is that the check in is automatic when the object is deleted. If you have to manually check in the license after deleting the object, keep line 48 as originally written.

src/ansys/dpf/core/server_context.py
```diff
@@ -38,6 +40,111 @@ def same_licensing_context(first, second):
...
+
+    Parameters
+    ----------
+    increment_name: str, optional
+         License increment to check out. To improve a script efficiency, this license increment
```
JennaPaikowsky, line 53:
```suggestion
         License increment to check out. To improve script efficiency, this license increment
```

src/ansys/dpf/core/server_context.py
```diff
@@ -38,6 +40,111 @@ def same_licensing_context(first, second):
...
+         should be consistent with the increments required by the following Operators. If ``None``,
+         the first available increment of this
+         `list <https://dpf.docs.pyansys.com/version/dev/user_guide/getting_started_with_dpf_server.
+         html#ansys-licensing>`_
+         is checked out.
+    license_timeout_in_seconds: float, optional
+         If no increment is available until then, check out fails. Defaults is:
```
JennaPaikowsky, line 60:
```suggestion
         If an increment is not available by the maximum time set here, check out fails. Default is:
```

src/ansys/dpf/core/server_context.py
```diff
@@ -38,6 +40,111 @@ def same_licensing_context(first, second):
...
+        Returns
+        -------
+        str
+        """
+        status_operator = dpf_operator.Operator("license_status", server=self._server)
+        return status_operator.eval()
+
+
 class ServerContext:
     """The context allows to choose which capabilities are available server side.
```
JennaPaikowsky, line 149:
```suggestion
    """The context allows you to choose which capabilities are available server side.
```

src/ansys/dpf/core/server_context.py
cbellot000: that's correct @JennaPaikowsky, thanks

src/ansys/dpf/core/runtime_config.py
```diff
@@ -182,3 +182,18 @@ def num_threads(self):
     @num_threads.setter
     def num_threads(self, value):
         self._data_tree.add(num_threads=int(value))
+
+    @property
+    def license_timeout_in_seconds(self):
+        """Sets the default number of threads to use for all operators,
+        default is omp_get_num_threads.
```
PProfizi, line 189:
```suggestion
        """Sets the default license timeout in seconds.
```

src/ansys/dpf/core/runtime_config.py
```diff
@@ -182,3 +182,18 @@ def num_threads(self):
...
+
+        Returns
+        -------
+        float
+        """
+        return self._data_tree.get_as("license_timeout_in_seconds", types.double)
+
+    @license_timeout_in_seconds.setter
+    def license_timeout_in_seconds(self, value):
```
PProfizi, line 198:
```suggestion
    def license_timeout_in_seconds(self, value):
        """Sets the license timeout in seconds."""
```

PProfizi (approved):

PR #849, opened 2023-03-22
Feat/preprate 2023.2.pre1 premium default

PProfizi:
Starting with `dpf_server_2023.2.pre1`, the licensing logic changes to an operator-based logic where each operator is tagged as either requiring a given (specific or any among a list) license to be checked-out, or as only requiring a license to exist.
This means the `ServerContext` being **Premium** means operators are **allowed** to check-out a license, and is the default behavior.
The `ServerContext` being **Entry** means operators are **not allowed** to check-out a license and will throw an error.

A DPF Server started with a **Premium** `ServerContext` cannot be set back to **Entry**.
Thus, the `server_context.py/set_default_server_context()` function should not try to apply an new default **Entry** context to an already running global server started as **Premium**.

For testing, having new servers being **Premium** by default means that:
- no more need to set the default context as **Premium** in the `conftest.py`.
- features to be tested as **Entry** should use a specific server started as **Entry**.

For examples:
- still need to mark the examples as requiring **Premium**
- no more need to set_default_server_context(Premium) in the first lines.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
     uses: ./.github/workflows/docs.yml
     with:
       ANSYS_VERSION: "232"
-      standalone_suffix: ${{ github.event.inputs.standalone_branch_suffix || '' }}
+      standalone_suffix: ${{ github.event.inputs.standalone_branch_suffix || '_test' }}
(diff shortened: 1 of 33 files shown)


PProfizi:
@cbellot000 working on setting Premium as default

PR #857, opened 2023-03-23
Update self.live variable when shutting down

ansys-akarcher:
self.live was updated for LegacyGrpcServer but was not for GrpcServer.

This causes issues when shutting down and starting a new server immediately from the same variable.

Perhaps it may be better to add a mechanism upper in the class hierarchy in BaseServer so this variable gets updated for every child class ?


cbellot000 (approved):

cbellot000:
@ansys-akarcher mayeb you could add the script you created to reproduce the bug as a unit test? they are in "/tests" folder

PProfizi (approved):

PR #873, opened 2023-03-30
Optimization of plotting methods: mesh_to_pyvista operator

rafacanton:
Plotting methods are optimized by means of the mesh_to_pyvista operator


rafacanton:
Tests are passing because for all server versions (both retro and current), the pyVista `UnstructuredGrid` is being generated at the `except` block. This PR should be merged after the `dpf_standalone` repo is updated with the server changes and then the tests and examples with the current version will use the new code

rafacanton:
@PProfizi This is ready to go (explicitly setting the server to run the operator as the same server of the mesh did the trick). The `CodeCov` is also in green (all the test in the latest version are covering the new functionality). The new doc also looks good (top new, bottom old):
image
image

PProfizi (approved):

PProfizi (approved):

PR #920, opened 2023-04-26
Fix memory issues in mesh_to_pyvista

rafacanton:


rafacanton:
Post pipelines are failing not due to the memory issue, but because we are not pulling the master branch (will be fixed in #918)

PR #935, opened 2023-05-04
Added method exposing operator_as_input connection

ansys-akarcher:
This new method finalizes the implementation of operator_connect_operator_as_input which was previously only available from the C API, and is now available from the gRPC API.

### DIFF
--- a/src/ansys/dpf/core/dpf_operator.py
+++ b/src/ansys/dpf/core/dpf_operator.py
@@ -267,6 +267,18 @@ class Operator:
             errormsg = f"input type {inpt.__class__} cannot be connected"
             raise TypeError(errormsg)
 
+    @version_requires("6.0")
+    def connect_operator_as_input(self, pin, op):
+        """Connects an operator as an input on a pin.
+        Parameters
+        ----------
+        pin : int
+            Number of the output pin. The default is ``0``.
+        op : :class:`ansys.dpf.core.dpf_operator.Operator`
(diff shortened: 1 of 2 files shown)


src/ansys/dpf/core/dpf_operator.py
```diff
@@ -267,6 +267,18 @@ def connect(self, pin, inpt, pin_out=0):
             errormsg = f"input type {inpt.__class__} cannot be connected"
             raise TypeError(errormsg)
 
+    @version_requires("6.0")
```
anslpa, line 270: 6.2

src/ansys/dpf/core/dpf_operator.py
```diff
@@ -267,6 +267,18 @@ def connect(self, pin, inpt, pin_out=0):
...
+    def connect_operator_as_input(self, pin, op):
+        """Connects an operator as an input on a pin.
+        Parameters
+        ----------
+        pin : int
+            Number of the output pin. The default is ``0``.
+        op : :class:`ansys.dpf.core.dpf_operator.Operator`
+            Requested type of the output. The default is ``None``.
+        """
+        self._api.operator_connect_operator_as_input(self, pin, op)
```
PProfizi, line 280: @ansys-akarcher hey, why would this line be marked as not-tested do you think?

PProfizi (approved):

PR #936, opened 2023-05-05
Update examples to use core.make_tmp_dir_server

PProfizi:
make_tmp_dir_server has recently been fixed to actually create the folder if it does not already exist.


anslpa (approved):

germa89:
I think `dpf.core.make_tmp_dir_server(dpf.SERVER)` line is a bit redundant, since if you are on local you should not needed (part of the logic), and if you are on remote, you should create it automatically (if needed) and the file server path if you needed it it is returned when you uploaded.

In PyMAPDL CICD that line is required to initialize `dpf.SERVER` otherwise it is `None`. This makes this line redundant or not very useful. Please notice we rely on the legacy grpc server because we use a DPF docker image.
I will see if I can update it these days. Any advice is welcomed.

PR #942, opened 2023-05-09
Fix/result info available qualifier labels

PProfizi:
Correcting a bug in `result_info.available_qualifier_labels`.
Cannot add a corresponding test in this PR as it requires a fix in `ansys-dpf-gate` which is not available without updating `dist/` in dpf-standalone 232 branches, which is not necessarily what we want.


tests/test_resultinfo.py
```diff
@@ -72,6 +71,10 @@ def test_print_result_info(model):
     print(model.metadata.result_info)
 
 
+def test_result_info_available_qualifier_labels(model):
+    assert model.metadata.result_info.available_qualifier_labels == []
+
+
```
PProfizi, line 77:
```suggestion
```
Remove test as it blocks the fix due to the CI still being on 232, and ansys-dpf-gate not being updated in dpf-standalone 232

rafacanton (approved):

Issue #960, opened 2023-05-22
Example on rescope


PProfizi:
### Description of the modifications

Add an example on rescope of results to get them ordered by node ID, element ID, or the same as a given mesh/scoping.

PR #961, opened 2023-05-22
Add rescope example

PProfizi:
Solves #960


examples/00-basic/13-data_reordering.py
```diff
@@ -0,0 +1,54 @@
...
+# The link between data position and corresponding entity ID is defined by the field's scoping.
+scoping = disp_field_0.scoping
+# This scoping is Nodal with several thousand entities (node IDs):
+print(scoping)
+# The first 10 node IDs in the scoping:
+print(scoping.ids[:10])
+# One can indeed see that the node IDs are not in ascending, descending or any particular order.
+
+# We can compare it to the order the mesh's nodes are in:
+nodes_scoping = model.metadata.meshed_region.nodes.scoping
```
rafacanton, line 36: @PProfizi From reading the example, it could inferred that the mesh scopings are _always_ ordered, which is not true. I mean, let's say that a user wants to have the data ordered by node Id. This example should address that. The solution offered in the example is to use the mesh scoping, but it can be the case that the mesh scoping is not ordered.

I would remark that the order of the mesh scoping is not mandatory. If a user wants a 100% ordered field, then the approach that works 100% of the times is:

```
reordered_fields = dpf.operators.scoping.rescope_fc(
    fields_container=fields,
    mesh_scoping= np.sort(fields[0].scoping.ids),
).outputs.fields_container()
```

With this approach we also make sure that we don't need to tackle the `default_value` pin of the `rescope` operator, as you are rescoping with exactly the same Ids you started with (not more, not less)

examples/00-basic/13-data_reordering.py
```diff
@@ -0,0 +1,54 @@
...
+
+# To improve performance, the result data comes ordered the same way it is stored on the server,
+# which is not necessarily by node or element ID.
+# The link between data position and corresponding entity ID is defined by the field's scoping.
+scoping = disp_field_0.scoping
+# This scoping is Nodal with several thousand entities (node IDs):
+print(scoping)
+# The first 10 node IDs in the scoping:
+print(scoping.ids[:10])
+# One can indeed see that the node IDs are not in ascending, descending or any particular order.
```
JennaPaikowsky, line 33:
```suggestion
# You can see that the node IDs are not in ascending, descending, or any particular order.
```

examples/00-basic/13-data_reordering.py
```diff
@@ -0,0 +1,54 @@
...
+
+# We can compare it to the order the mesh's nodes are in:
+nodes_scoping = model.metadata.meshed_region.nodes.scoping
+print(nodes_scoping)
+print(nodes_scoping.ids[:10])
+
+# The mesh's node scoping is in this case in ascending order.
```
JennaPaikowsky, line 40:
```suggestion
# The mesh's node scoping is, in this case, in ascending order.
```

PR #965, opened 2023-05-30
Fix find_ansys method to return installed ansys-dpf-server in priority to local Ansys installs

PProfizi:
Fix find_ansys logic to return the latest installed ansys-dpf-server package first instead of the latest including Ansys installations

### DIFF
--- a/src/ansys/dpf/core/misc.py
+++ b/src/ansys/dpf/core/misc.py
@@ -80,9 +80,7 @@ def get_ansys_path(ansys_path=None):
     # First check the environment variable for a custom path
     if ansys_path is None:
         ansys_path = os.environ.get("ANSYS_DPF_PATH")
-    # Then check for usual installation folders with AWP_ROOT and find_ansys
-    if ansys_path is None:
-        ansys_path = os.environ.get("AWP_ROOT" + __ansys_version__)
+    # Then check among installed packages, AWP_ROOT variables, and usual installation folders
     if ansys_path is None:
         ansys_path = find_ansys()
     # If still no install has been found, throw an exception
@@ -117,8 +115,13 @@ def _find_latest_ansys_versions():
             installed_packages_list[
                 packaging.version.parse(_pythonize_awp_version(awp_version))
             ] = ansys_path
+    if len(installed_packages_list) > 0:
+        return installed_packages_list[sorted(installed_packages_list)[-1]]
 
+
+def _find_latest_ansys_dpf_server():
     installed_packages = pkg_resources.working_set
+    installed_packages_list = {}
     for i in installed_packages:
         if "ansys-dpf-server" in i.key:
             file_name = pkg_resources.to_filename(i.project_name.replace("ansys-dpf-", ""))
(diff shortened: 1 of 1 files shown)


src/ansys/dpf/core/misc.py
```diff
@@ -80,9 +80,7 @@ def get_ansys_path(ansys_path=None):
     # First check the environment variable for a custom path
     if ansys_path is None:
         ansys_path = os.environ.get("ANSYS_DPF_PATH")
-    # Then check for usual installation folders with AWP_ROOT and find_ansys
-    if ansys_path is None:
-        ansys_path = os.environ.get("AWP_ROOT" + __ansys_version__)
+    # Then check among installed packages, AWP_ROOT variables, and usual installation folders
     if ansys_path is None:
         ansys_path = find_ansys()
```
cbellot000, line 85: @PProfizi is `find_ansys()` doing ` ansys_path = os.environ.get("AWP_ROOT" + __ansys_version__)`?

src/ansys/dpf/core/misc.py
```diff
@@ -156,6 +161,9 @@ def find_ansys():
     >>> path = find_ansys()
 
     """
+    latest_server = _find_latest_ansys_dpf_server()
```
cbellot000, line 164: @PProfizi I don't think that it's the expected behavior. We want the latest version through both the standalone and the ansys versions. That's why it was a single function

PProfizi: @cbellot000 I think it is. Otherwise, what you got locally and we called a bug, is actually the intended behavior:
- we have ansys-dpf-server-2024.1.pre0 installed
- we have AWP_ROOT241, which is translated to 2024.1.0
- > the AWP_ROOT241 wins

Yet, the behavior above is not what we want IMO.

src/ansys/dpf/core/misc.py
PProfizi: @cbellot000, no, as this was the actual issue.
If you prioritize `AWP_ROOT+__ansys_version__`, then you do not use the standalone installed in your venv.

With the new implementation, `find_ansys` goes through all installed `ansys-dpf-server` and takes the latest if any, then goes through `AWP_ROOT` variables and takes the latest if any, which results in the same thing as prioritizing `__ansys_version` which is the latest anyway. This is actually one thing less to maintain. Or then I am missing a use-case

src/ansys/dpf/core/misc.py
rafacanton: @cbellot000 I agree with what @PProfizi is saying. If a user installs some server version (say 2023.pre.1) in the venv and then 1 year later an Ansys unified install is installed (say 25.2), and the user wants to use this one instead, what he should do is uninstall the server. The problem is how to communicate this (I guess that printing the ansys path to the std output each time we import dpf is not a good thing)

cbellot000: @rafacanton and @PProfizi let's discuss this in a meeting

PR #984, opened 2023-06-12
Add volume rendering

PProfizi:
Enable plotting a volume rendering of a field


AlejandroFernandezLuces (changes requested):
The functionality seems good to me, nice work 🙂 . I've tried this with the example you provided offline and it achieves what you were looking for. In case you need animations, as in the video you provided in the description, I can help you with that.

Before merging, please perform an unit test for this function. You can use [pytest-pyvista](https://github.com/pyvista/pytest-pyvista) plugin to do so. 

Also, it would be nice if you could upload an usage example. This would greatly improve the exposure and usability of this feature, up to you.

Issue #967, opened 2023-05-31
Crash in min_max_over_time_by_entity


janvonrickenbach:
### Before submitting the issue

- [X] I have checked for [Compatibility](https://dpf.docs.pyansys.com/version/stable/getting_started/compatibility.html#id1) issues
- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

Connect to a dpf server and run the code below. It crashes on the last line. I tried with a server in a docker container as well as the server from the current installer (24R1). Other min_max operators with the same input data work without problems (e.g. min_max_over_label_fc)
```
    test_field = dpf.Field(nature=natures.scalar)
    test_field.append([1.], 1)
    test_field.append([5.], 2)
    test_field.append([3.], 3)

    test_field2 = dpf.Field(nature=natures.scalar)
    test_field2.append([0.], 1)
    test_field2.append([5.], 2)
    test_field2.append([6.], 3)

    fields_container = dpf.FieldsContainer()
    fields_container.add_label("time")
    fields_container.add_field({"time": 1}, test_field)
    fields_container.add_field({"time": 2}, test_field2)

    max_op = dpf.operators.min_max.min_max_over_time_by_entity()
    max_op.inputs.fields_container.connect(fields_container)
    max = max_op.outputs.max()
```

### Steps To Reproduce

See description

### Which Operating System causes the issue?

Windows

### Which DPF/Ansys version are you using?

DPF Server 2023.2.pre1

### Which Python version causes the issue?

3.10

### Installed packages

accessible-pygments           0.0.4
alabaster                     0.7.13
ansys-dpf-composites          0.2.dev0  D:\ANSYSDev\pydpf-composites
ansys-dpf-core                0.8.1
ansys-dpf-gate                0.3.1
ansys-dpf-gatebin             0.3.1
ansys-grpc-dpf                0.7.1
ansys-sphinx-theme            0.9.8                                 
anyio                         3.6.2                                 
appdirs                       1.4.4                                 
argon2-cffi                   21.3.0                                
argon2-cffi-bindings          21.2.0                                
arrow                         1.2.3                                 
astroid                       2.11.7                                
asttokens                     2.2.1                                 
async-lru                     2.0.2                                 
attrs                         23.1.0                                
Babel                         2.12.1
backcall                      0.2.0
beautifulsoup4                4.12.2
bleach                        6.0.0
build                         0.8.0
cachetools                    5.3.0
certifi                       2022.12.7
[...]

PR #996, opened 2023-06-14
Fixed handling of empty Time Freq Support

ansys-akarcher:
These changes solve the #967 crash, when using fields_container_factory.over_time_freq_fields_container(...)

Edit: this code is wrong.
we should use a dict when we want a TimeFreqSupport
And min_max_* should not be dependent on TimeFreqSupport


ansys-akarcher:
I reverted as previously said,

Also the following code was enough to produce a crash:

```python
fields_container = dpf.FieldsContainer()
print(fields_container.time_freq_support)
```
```
Traceback (most recent call last):
  File "C:\Users\akarcher\pydpf-core-venv\bug_minmax_overtime\test.py", line 16, in <module>
    print(fields_container.time_freq_support)
  File "C:\Users\akarcher\pydpf-core-venv\pydpf-core\src\ansys\dpf\core\fields_container.py", line 427, in time_freq_support
    return self._get_time_freq_support()
  File "C:\Users\akarcher\pydpf-core-venv\pydpf-core\src\ansys\dpf\core\collection.py", line 397, in _get_time_freq_support
    time_freq = support_api.support_get_as_time_freq_support(support)
  File "d:\ansysdev\ansys-dpf-pygate\ansys-dpf-gate\ansys\dpf\gate\generated\support_capi.py", line 71, in support_get_as_time_freq_support
    raise errors.DPFServerException(sError.value)
ansys.dpf.gate.errors.DPFServerException: Provided shared object is not a FieldSupport !
```

cbellot000 (approved):

ansys-akarcher:
The new test "test_fields_container_empty_tf_support()" doesn't work with LegacyGrpcServer, not sure on how to proceed from there, as the grpc interface returns an UNAVAILABLE when retrieving a nullptr (which makes sense)

cbellot000 (approved):

cbellot000 (approved):

PProfizi (approved):

ansys-akarcher:
Completes #967

PR #1003, opened 2023-06-22
Feat/datasources namespace

cbellot000:


tests/test_datasources.py
```diff
@@ -89,3 +89,15 @@ def test_delete_auto_data_sources(server_type):
 
     gc.collect()
     assert ref() is None
+
+
+@conftest.raises_for_servers_version_under("7.0")
+def test_register_namespace(allkindofcomplexity, server_type):
+    data_sources = dpf.core.DataSources(server=server_type)
+    data_sources.set_result_file_path(allkindofcomplexity)
```
jose-henriquezroa, line 97: @cbellot000 So the result_key & namespace are set internally depending on which file is given ?

jose-henriquezroa (approved):

tests/test_datasources.py
cbellot000: sorry I don't understand the question @jose-henriquezroa

jose-henriquezroa: @cbellot000 because I see that the displacement source works without manually calling "register_namespace". Then it fails once a bogus namespace is set (i.e notmapdl). So it means there some type of mapping (file type) -> (key, namespace) somewhere in dpf. Which I think, thinking back, you showed it to me somewhere along the line.

Sorry for the question, source operators are still something that I haven't worked with on dpf.

PR #1025, opened 2023-07-12
add data-tree list-attributes

jose-henriquezroa:
Exposed DataTree function for listing attributes.

### DIFF
--- a/src/ansys/dpf/core/data_tree.py
+++ b/src/ansys/dpf/core/data_tree.py
@@ -482,6 +482,14 @@ class DataTree:
             out = DataTree(data_tree=obj, server=self._server)
         return out
 
+    def get_attributes(self):
+        coll_obj = collection.StringCollection(
+            collection=self._api.dpf_data_tree_get_available_attributes_names_in_string_collection(self),
+            server=self._server,
+        )
+
+        return coll_obj.get_integral_entries()
+
     def __setattr__(self, key, value):
         if key == "_common_keys" or key in self._common_keys:
             return super.__setattr__(self, key, value)
--- a/tests/test_data_tree.py
+++ b/tests/test_data_tree.py
@@ -316,3 +316,18 @@ def test_unsupported_types_data_tree(server_type):
         data_tree.add(data1=[[1]])
     with pytest.raises(TypeError):
         data_tree.add(data1=(1, 2))
+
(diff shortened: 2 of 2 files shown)


src/ansys/dpf/core/data_tree.py
```diff
@@ -482,6 +482,16 @@ def get_as(self, name, type_to_return=types.string):
             out = DataTree(data_tree=obj, server=self._server)
         return out
 
+    def get_attributes(self):
```
cbellot000, line 485: @jose-henriquezroa could you please add a dosctring and in this docstring precise whether sub attributes are returned or not (whether this function is recursive basically?))

To comply with our other python classes I would make this function a property instead, what do you think?

jose-henriquezroa: @cbellot000 I will add the docstring and add an option for recursively fetching attributes from subtrees as well. I think the option to do that will be very useful indeed.

src/ansys/dpf/core/data_tree.py
```diff
@@ -482,6 +482,91 @@ def get_as(self, name, type_to_return=types.string):
...
+
+        for sub_tree_name in self.get_sub_tree_names():
+            sub_tree = self.get_as(sub_tree_name, types.data_tree)
+            sub_dic = {}
+            sub_tree.__to_dict(sub_dic)
+            dic[sub_tree_name] = sub_dic
+
+    def to_dict(self):
+        """
+        Returns a dictionary representation of the DataTree
```
cbellot000, line 549: I like the idea!
```suggestion
        Returns a read only dictionary representation of the DataTree
```

src/ansys/dpf/core/data_tree.py
```diff
@@ -482,6 +482,91 @@ def get_as(self, name, type_to_return=types.string):
...
+        coll_obj = collection.StringCollection(
+            collection=self._api.dpf_data_tree_get_available_attributes_names_in_string_collection(
+                self
+            ),
+            server=self._server,
+        )
+
+        return coll_obj.get_integral_entries()
+
+    def get_sub_tree_names(self):
```
cbellot000, line 510: Could we use properties here please to be consistent with other classes, ie: "attribute_name", "sub_stree_names"

cbellot000 (approved):

PR #1029, opened 2023-07-13
New API for MeshInfo

rafacanton:
- Simplification of API for the `MeshInfo` (no need to pass the type of a property to retrieve it)
- Elimination of the `body_name`, `zone_name` and `topology` scenarios of the MeshInfo tests to avoid failing pipelines while doing the next server migration (will be reverted)


tests/test_mesh_info.py
```diff
@@ -159,33 +159,33 @@ def test_output_mesh_info_provider_fluent(server_clayer):
...
     assert num_cells == 1344
     assert num_faces == 2773
     assert num_nodes == 1430
 
     # """************************ BODIES ************************"""
-
-    # """************ Name ************"""
-    body_names = mesh_info_out.get_property("body_name", dpf.StringField)
+    """
+    # ************ Name ************
```
PProfizi, line 172:
```suggestion
    # ************ Name ************
```

PProfizi (approved):

PR #1036, opened 2023-07-18
Add "plot streamlines" capabilities

anslpa:
## Enhancement

Add streamlines plotting capabilities to the DpfPlotter. 

## Work done

- add "helpers" module in ansys.dpf.core, with a first "utils" and "streamlines" submodules. 
- add a "compute_streamlines" method in the helpers module, with:
   - capability to return source (source is displayed with this argument)
   - all the kwargs of "pyvista.DataSetFilters.streamlines" and "pyvista.DataSetFilters.streamlines_from_source"
- add "add_streamline" method with: 
    - capability to change the radius of the streamline when plotting
    - a permissive option to allow the plot (to see the source location) even if the streamline is empty
- add examples for computing and plotting 2D and 3D models
- add "plane" capability for "show_figure" method (allows to plot in a defined plane, e.g. (XY))
- performance testing (see report below)
- tested with PyVista 0.38.1 and 0.40.*

## Performances
image

## Streamline plot example
image


src/ansys/dpf/core/plotter.py
```diff
@@ -567,6 +690,8 @@ def show_figure(self, **kwargs):
         **kwargs : optional
             Additional keyword arguments for the plotter. More information
             are available at :func:`pyvista.plot`.
+            "show_plane" argument can be used to specify the plan to view. For
```
rafacanton, line 693: plane

anslpa: Done

examples/06-plotting/08_plot_3d_streamlines.py
```diff
@@ -0,0 +1,172 @@
...
+# Get the velocity result at nodes:
+velocity_op = m_fluent.results.velocity()
+fc = velocity_op.outputs.fields_container()
+field = dpf.operators.averaging.to_nodal_fc(fields_container=fc).outputs.fields_container()[0]
+
+###############################################################################
+# Plot the streamlines adjusting the request
+# ------------------------------------------
+# The following steps show you how to create streamlines using DpfPlotter, with several sets
+# of parameters. It shows the issues that can happen, the adjustments that can be done.
```
JennaPaikowsky, line 53:
```suggestion
# of parameters. It demonstrates the issues that can happen and the adjustments that you can make.
```

examples/06-plotting/08_plot_3d_streamlines.py
```diff
@@ -0,0 +1,172 @@
...
+
+# First, a DpfPlotter is created and the streamline is created with default values.
+pl0 = DpfPlotter()
+try:
+    pl0.add_mesh(meshed_region=meshed_region, opacity=0.3)
+    pl0.add_streamlines(meshed_region=meshed_region, field=field)
+    pl0.show_figure(show_axes=True)
+except:
+    # It throws and ends here, because source point are not correctly set.
```
JennaPaikowsky, line 62:
```suggestion
    # It throws an error and ends here, because source points are not set correctly.
```

examples/06-plotting/08_plot_3d_streamlines.py
```diff
@@ -0,0 +1,172 @@
...
+    # They don't match with the mesh coordinates velocity values
+    # are known for.
```
JennaPaikowsky, line 64: That are known?

examples/06-plotting/08_plot_3d_streamlines.py
```diff
@@ -0,0 +1,172 @@
...
+    # To fix this issue, the source_center parameter is used to move
+    # the streamlines source center.
+    pass
+
+# Then, using "source_center" argument that moves the source center,
+# the "return_source" arguments that displays the source, and
+# "permissive" option that allows the display of the source even if the computed
+# streamline size is zero, the source coordinates can be correctly set:
```
JennaPaikowsky, line 72:
```suggestion
# Then, you can correctly set the source coordinates using the "source_center" argument that moves the source center,
# the "return_source" arguments that displays the source, and
# "permissive" option that allows you to display the source even, if the computed
# streamline size is zero.
```

examples/06-plotting/08_plot_3d_streamlines.py
```diff
@@ -0,0 +1,172 @@
...
+    source_center=(0.1, 0.1, 0.2),
+    permissive=True,
+)
+pl1.show_figure(show_axes=True)
+
+# After adjustment, correct values for "source_center" argument is set.
+# The "permissive" option can be removed.
+# Velocity data with small opacity to avoid to hide the streamlines can be displayed.
+# More settings are added to adapt the streamlines creation to the geometry and
+# the data of the model:
```
JennaPaikowsky, line 88:
```suggestion
# After the adjustment, the correct values for the "source_center" argument are set.
# You can remove the "permissive" option.
# You can display velocity data with a small opacity value to avoid hiding the streamlines.
# More settings are added to adapt the streamlines creation to the geometry and
# the data of the model:
```

examples/06-plotting/09_plot_surface_streamlines.py
```diff
@@ -0,0 +1,81 @@
+"""
+.. _plot_surf_streamlines:
+
+Plot 2D streamlines
+~~~~~~~~~~~~~~~~~~~
+This example shows how to plot streamlines of fluid simulation results,
```
JennaPaikowsky, line 6:
```suggestion
This example shows you how to plot streamlines of fluid simulation results,
```

src/ansys/dpf/core/plotter.py
```diff
@@ -496,6 +552,73 @@ def add_mesh(self, meshed_region, deform_by=None, scale_factor=1.0, **kwargs):
...
+        ----------
+        meshed_region : MeshedRegion
+            MeshedRegion the streamline will be computed on.
+        field : Field
+            Field containing raw vector data the streamline is
+            computed from. The data location must be nodal, velocity
+            values must be defined at nodes.
+        **kwargs : optional
+            Additional keyword arguments for the plotter. More information
+            are available at :func:`pyvista.DataSetFilters.streamlines`.
```
JennaPaikowsky, line 579:
```suggestion
            is available at :func:`pyvista.DataSetFilters.streamlines`.
```

JennaPaikowsky (approved):
Just a few comments

examples/06-plotting/08_plot_3d_streamlines.py
```diff
@@ -0,0 +1,171 @@
...
+# Plot the streamlines adjusting the request
+# ------------------------------------------
+# The following steps show you how to create streamlines using DpfPlotter, with several sets
+# of parameters. It demonstrates the issues that can happen and the adjustments that you can make.
+
+# First, a DpfPlotter is created and the streamline is created with default values.
+pl0 = DpfPlotter()
+try:
+    pl0.add_mesh(meshed_region=meshed_region, opacity=0.3)
+    pl0.add_streamlines(meshed_region=meshed_region, field=field)
```
cbellot000, line 59: @anslpa is the goal of this feature to plot streamline or to compute streamlines? We will not be able to retrieve the actual data ?

anslpa: Create fluid_helpers for streamline computation, either create a function or a class that would look like an operator. This method would be called "fluid_helpers.compute_streamline"

anslpa: Ideal would be to hav ea streamline per field in the fields container ? How the data is containerized per streamline in PolyData ?

anslpa: # PyDpf-Core
## Compute streamline and then plot it
streamlines_fc_3D = dpf.helpers.compute_streamline(meshed_region, field=velocity_field, source_center=(...), source_radius=0.001)
streamlines_fc_2D = dpf.helpers.compute_streamline(meshed_region=meshed_region, field=field, return_source=True, radius=0.00002, start_position=(0.005, 0.0005, 0.0), surface_streamlines=True)
pl2 = DpfPlotter()
pl2.add_streamlines(computed_streamline=streamlines_fc_3D, radius=0.001)
pl2.show_figure(show_axes=True)

## Plot streamline

```
pl2 = DpfPlotter()
pl2.add_streamlines(
    meshed_region=meshed_region,
    field=field,
    return_source=True,
    source_center=(0.56, 0.48, 0.0),
    radius=0.001,
    n_points=10,
    source_radius=0.075,
    max_time=10.0,
)
pl2.show_figure(show_axes=True)
```

# PyDpf-Post

PProfizi: @anslpa yes with https://docs.pyvista.org/version/stable/api/core/_autosummary/pyvista.DataSetFilters.streamlines_from_source.html#pyvista.DataSetFilters.streamlines_from_source
So the method taking in a PyDPF-Core geometry object would need to translate it to a `pv.DataSet`

anslpa: quid "return_source" ? source as field ?

anslpa: Handled in this PR: https://github.com/ansys/pydpf-core/pull/1062/files.

examples/12-streamlines/00_plot_3d_streamlines.py
rafacanton, line 1: @anslpa The folder should be `13-streamlines`, as 12 is now taken for `12-fluids`

anslpa: Updated!

examples/13-streamlines/00_plot_3d_streamlines.py
```diff
@@ -0,0 +1,207 @@
...
+###############################################################################
+# First, a DpfPlotter is created and the streamline is created with default values.
+pl0 = DpfPlotter()
+try:
+    streamline_obj = compute_streamlines(meshed_region=meshed_region, field=field)
+    pl0.add_mesh(meshed_region=meshed_region, opacity=0.3)
+    pl0.add_streamlines(streamlines=streamline_obj)
+    pl0.show_figure(show_axes=True)
+except:
+    # It throws an error and ends here, because source points are not set correctly.
```
cbellot000, line 68: @anslpa do we need the try catch in the example?

src/ansys/dpf/core/helpers/streamlines.py
```diff
@@ -0,0 +1,182 @@
...
+    object scripting with `ansys-dpf-core`.
+
+    """
+
+    def __init__(self, data):
+        super().__init__(data=data)
+
+
+def compute_streamlines(meshed_region, field, **kwargs):
+    """Compute the streamlines for a given mesh and velocity
```
cbellot000, line 92: will we be able to put a sphere object here?

examples/13-streamlines/00_plot_3d_streamlines.py
```diff
@@ -0,0 +1,207 @@
...
+###############################################################################
+# After the adjustment, the correct values for the "source_center" argument are set.
+# You can remove the "permissive" option.
+# You can display velocity data with a small opacity value to avoid hiding the streamlines.
+# More settings are added to adapt the streamlines creation to the geometry and
+# the data of the model:
+# - radius: streamlines radius
+# - n_points: source number of points
+# - source_radius
+# - max_time: allows the streamline to be computed along a certain length
```
PProfizi, line 103: @anslpa how is this "time" linked to a length value? How do I know what max_time to put?

src/ansys/dpf/core/helpers/streamlines.py
anslpa: @cbellot000 for now this is not possible to pass a "source" of any kind (even pyvista object) to compute the streamline. Computing streamlines from a source using "compute_streamlines" from a pyvista object source should be possible with some (small) changes. As far as we have that available, we could use the "helpers.streamlines.StreamlinesSource" object to handle the Geometry objects (however I am seeing that currently, only the "Line" and "Point" exists there). 

Is it needed to have the "compute_streamline" from a specific source changes in that PR ? Or can that be done later ?

examples/13-streamlines/00_plot_3d_streamlines.py
anslpa: @cbellot000 I am assuming that a lot of people will face the issue that I faced, which was that the source was defined outside of the mesh geometry, and when trying to plot in this case, it is throwing. This try/except intends to demonstrate what's not expected to work. Is that an issue ?

src/ansys/dpf/core/helpers/streamlines.py
PProfizi: @anslpa @cbellot000 I guess the idea is to add an optional "source" argument in the future which will take in a geometry object

examples/13-streamlines/00_plot_3d_streamlines.py
anslpa: @rafacanton, could you please help me to make that clearer ?

src/ansys/dpf/core/helpers/streamlines.py
```diff
@@ -0,0 +1,182 @@
...
+            raise AttributeError(
+                "streamlines must be a pyvista.PolyData or a dpf.FieldsContainer instance."
+            )
+
+    def _pv_data_set_to_fc(self):
+        """Convert pyvista.PolyData into FieldsContainer."""
+        raise Exception("Not implemented yet")
+
+    def _fc_to_pv_data_set(self):
+        """Convert pyvista.PolyData into FieldsContainer."""
```
MichaelNale, line 46: isn't it the other way round?

examples/13-streamlines/00_plot_3d_streamlines.py
PProfizi: @anslpa can't we catch the error ourselves when trying to plot an empty streamline? in `add_streamlines`

src/ansys/dpf/core/helpers/streamlines.py
anslpa: Good catch, it's fixed. Thanks!

src/ansys/dpf/core/helpers/streamlines.py
anslpa: @PProfizi yes, or create a "compute_streamlines_from_source" method as PyVista did

examples/13-streamlines/00_plot_3d_streamlines.py
rafacanton: @PProfizi Because streamlines are the first integrals of velocity. What we are doing is solving for this system of 3 ODEs:

```
dx/dt = v_x(x,y,z)
dy/dt = v_y(x,y,z)
dz/dt = v_z(x,y,z)
```

for an initial condition (x0,y0,z0). This integration will finish if:

- v->0 (the default value for this according to pyVista doc is 1e-12.
- the streamline leaves the domain.
- You integrate for a certain amount of time. The max_time controls this. If it is too short, the streamñines will be very short

rafacanton: So I would maybe rephrase as:

- max_time: maximum integration time of the streamline. It controls the streamline length.

examples/13-streamlines/00_plot_3d_streamlines.py
cbellot000: I think @PProfizi is right, and the issue you encountered should be added in the docstring but using try except in an example (failing secretly BTW) makes it feel like it's a randomly failing example

anslpa: As discussed, I'll set the "persmissive" default value to True and remove this "try/except".

examples/13-streamlines/00_plot_3d_streamlines.py
anslpa: Done, thank you!

PProfizi: @rafacanton ah ok, my mistake was thinking that a streamline is computed on a snapshot velocity field at a specific instant, hence no time. However moving along this snapshot flow is still equivalent to moving in "time" like a particle would.

PProfizi (approved):

PR #1042, opened 2023-07-19
Update DataTree

jose-henriquezroa:


src/ansys/dpf/core/data_tree.py
```diff
@@ -535,18 +537,18 @@ def get_sub_tree_names(self):
...
             dic[attribute_name] = self.get_as(attribute_name)
 
-        for sub_tree_name in self.get_sub_tree_names():
+        for sub_tree_name in self.get_sub_tree_names:
             sub_tree = self.get_as(sub_tree_name, types.data_tree)
             sub_dic = {}
             sub_tree.__to_dict(sub_dic)
             dic[sub_tree_name] = sub_dic
 
     def to_dict(self):
```
PProfizi, line 549:
```suggestion
    @property
    def dict(self):
```
You can make this a property
You could also define the `__dict__` dunder method so that `dict(your_object)` works

src/ansys/dpf/core/data_tree.py
```diff
@@ -535,18 +537,18 @@ def get_sub_tree_names(self):
...
         """
-        Returns a dictionary representation of the DataTree
```
PProfizi, line 549: How do you enforce the fact it is read-only if you return an actual Python `dict`?

src/ansys/dpf/core/data_tree.py
```diff
@@ -507,6 +508,7 @@ def get_attribute_names(self):
 
         return coll_obj.get_integral_entries()
 
+    @property
     def get_sub_tree_names(self):
```
PProfizi, line 512:
```suggestion
    def sub_tree_names(self):
```

src/ansys/dpf/core/data_tree.py
```diff
@@ -482,6 +482,7 @@ def get_as(self, name, type_to_return=types.string):
             out = DataTree(data_tree=obj, server=self._server)
         return out
 
+    @property
     def get_attribute_names(self):
```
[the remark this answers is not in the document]
PProfizi: Also above, why not just call the methods `to_json` or `from_txt` without the `read` and `write` parts. This too seems explicit enough without it.

src/ansys/dpf/core/data_tree.py
PProfizi: Also, I'd personally integrate `__to_dict` in the `dict` property, and then call directly the `dict` property in other parts of the code.

jose-henriquezroa: @PProfizi Will do! Wasn't aware of this!

src/ansys/dpf/core/data_tree.py
jose-henriquezroa: It was originally a comment from @cbellot000 meaning that modifying the attributes & attribute-names in the dict didn't actually modify anything the DataTree. Which on second thought, the fact that we return a dict makes it clear enough. But no harm in leaving there I think.

src/ansys/dpf/core/data_tree.py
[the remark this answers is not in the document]
PProfizi: @jose-henriquezroa you do not have to implement the setter and make it raise an error if what you want is for people to not be able to set. Simply not specifying a setter is enough.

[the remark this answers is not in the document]
jose-henriquezroa: @PProfizi You're right, my mistake. I misunderstood  the "property". 

Nonetheless, in the very particular case of the DataTree we've also defined the __setattr__ function. Which means that if the user does data_tree.get_attribute_name = "hello", instead of throwing an error it's going to define an attribute in the data_tree object with key "get_attribute_name" and value "hello". So I guess using "properties" is too risky here still.

cbellot000 (approved):

PProfizi (approved):

PR #1045, opened 2023-07-20
Fix issues with PyVista 0.41.0

PProfizi:
Hotfix for PyVista 0.41.0 released this morning.

This apparently removed the deprecated `pv.rcParams` we were still using and is breaking all our pipeline runs.

Still having an issue (which I never managed to reproduce locally neither on Windows not on Linux) with doc generation on GitHub though, requiring `pyvista==0.36.1`:

image


rafacanton (approved):

jose-henriquezroa (approved):

germa89:
This is great. But we will need a patch release for this ;)

PR #1072, opened 2023-07-26
Documentation: update availability information of Custom Operators section

anslpa:


cbellot000 (approved):

anslpa:
> @anslpa are we really going to have to update that every time? Can't we find a way to explain this without having to update at each release?

This is not an "update", this is a fix (python operators where available from 23.1, not from 22.1)

janvonrickenbach:
@anslpa I think this should be fixed for older versions of the documentation (0.7, 0.6 etc) as well. Currently I see this only fixed for the "dev" version of the documentation.

PR #1077, opened 2023-07-27
Streamlines: access streamline data as FieldsContainer

anslpa:


src/ansys/dpf/core/helpers/streamlines.py
```diff
@@ -40,11 +45,110 @@ def __init__(self, data):
...
+        npoints = len(points)
+        vpoly = vtk.vtkPolyData()
+        vpoints = vtk.vtkPoints()
+        vpoints.SetNumberOfPoints(npoints)
+        vtk_array_points = pv.convert_array(arr=points)
+        vpoints.SetData(vtk_array_points)
+        vpoly.SetPoints(vpoints)
+        vcells = vtk.vtkCellArray()
+        for i in range(0, ncells):
+            this_cell_points = cell_points[i]
```
anslpa, line 136: Maybe this could be improved using a method that sets the whole connectivity data array and the whole cell type array, however I did not find it (yet ?).

src/ansys/dpf/core/helpers/streamlines.py
```diff
@@ -28,44 +33,153 @@ def __init__(self, data):
...
+            cell_type_id = cell_types[i]
+            if cell_type_id == element_types.Line2.value:
+                vtk_cell = vtk.vtkLine()
+            elif cell_type_id == element_types.EMagLine.value:
+                vtk_cell = vtk.vtkPolyLine()
+            elif element_types.Unknown.value:
+                vtk_cell = vtk.vtkGenericCell()
```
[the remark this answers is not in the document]
anslpa: Looks ok

examples/13-streamlines/00_plot_3d_streamlines.py
```diff
@@ -79,6 +79,14 @@
 )
 pl1.show_figure(show_axes=True)
 
+###############################################################################
+# Access the streamlines data
+# ---------------------------
+# The Streamlines data can be requested as a Field.
+# The streamlines Field has a MeshedRegion as support.
+streamlines_field = streamline_obj.as_field()
+streamlines_data = streamlines_field.data
```
anslpa, line 88: For 2D examples, manipulate return_cpos and cpos to have a correct view

examples/13-streamlines/00_plot_3d_streamlines.py
```diff
@@ -79,6 +79,14 @@
 )
 pl1.show_figure(show_axes=True)
 
+###############################################################################
```
anslpa, line 82: Look the following commit to see how I tried to get the animation working for streamlines: https://github.com/ansys/pydpf-core/pull/1077/commits/90f9ef33ac5317aa5020b6afbc14873a8813b6cc. It's not working because it look either there's a bug in the "add_field" method or in the streamlines Field object.

src/ansys/dpf/core/helpers/streamlines.py
```diff
@@ -4,8 +4,9 @@
 import numpy as np
 import warnings
 
+from ansys.dpf import core as dpf
```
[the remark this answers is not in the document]
PProfizi: @anslpa I would say yes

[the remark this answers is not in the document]
anslpa: I created an item, that this can be done later, it won't be done in this PR.

src/ansys/dpf/core/helpers/streamlines.py
```diff
@@ -28,44 +29,173 @@ def __init__(self, data):
...
+            self._vtk_type_from_dpf_type(c, cell_types_converted, vtk)
+        nodes_scoping = dpf.Scoping(location=locations.nodal, server=server)
+        nodes_scoping.ids = np.arange(1, data_set.n_points + 1)
+        streamlines_field = dpf.Field(location=locations.nodal, server=server)
+        fdef = streamlines_field.field_definition
+        if len(array_names) > 0:
+            fdef.name = array_names[0]
+        streamlines_field.scoping = nodes_scoping
+        if len(data_arrays) > 0:
+            streamlines_field.data = data_arrays[0]
```
rafacanton, line 96: @anslpa If I understand correctly, what you are doing here is first doing a loop in all points in `self._pv_data_set` regardless of which streamline in particular they belong to, and then putting all of them in a very big field. What I would suggest is creating a `FieldsContainer`with one Field per streamline with label "streamline" (maybe?), and each one having a mesh support as the line mesh of the streamline. I expect that when we ship this as an operator it will we done in this way, as computing each streamline is independent from the other and all of them cuold be computed in parallel in a pragma omp for.

anslpa: After discussion with Rafael, we decided to return a FieldContainer with the following label : "line". Each line of the streamlines would be defined by a field. Each field would have a MeshedRegion associated with one cell element (multi lines cell) and all the connected points. 

Later, the "multi lines cell" needs to be handled by the PyVista plotter when running mesh_single_cell_multi_lines.plot()" (this might imply changes in the way we give the cells types to pyvista, in _vtk_helpers).

PR #1084, opened 2023-07-31
Use cpos argument to select a plane for plot

PProfizi:
instead of creating a new "plane" argument

### DIFF
--- a/src/ansys/dpf/core/plotter.py
+++ b/src/ansys/dpf/core/plotter.py
@@ -355,11 +355,6 @@ class _PyVistaPlotter:
             self._plotter.camera_position = cpos
 
         # Show depending on return_cpos option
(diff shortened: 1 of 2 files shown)


rafacanton (approved):

src/ansys/dpf/core/plotter.py
```diff
@@ -355,11 +355,6 @@ def show_figure(self, **kwargs):
             self._plotter.camera_position = cpos
 
         # Show depending on return_cpos option
-        plane = kwargs.pop("plane", None)
```
anslpa, line 358: @PProfizi it looks good to me, thank you ! It would be great if we could also update cpos to have the image zoomed, and an explanation of what has been done to get the corresponding cpos.

anslpa (approved):
Approved with suggestion!

PR #1123, opened 2023-08-16
Remove all result files still packaged with ansys-dpf-core

PProfizi:
Make all examples downloadable.
Do not package any result files within ansys-dpf-core anymore.


cbellot000:
@PProfizi could you please give the size of the package before and after?

src/ansys/dpf/core/examples/examples.py
```diff
@@ -13,6 +13,7 @@
 from ansys.dpf.core.core import upload_file_in_tmp_folder
 from ansys.dpf.core import path_utilities
 from ansys.dpf.core import DataSources
+from ansys.dpf.core.examples.downloads import _download_file
```
cbellot000, line 16: @PProfizi what about lines 25 to 35 of this file?

PR #1139, opened 2023-09-07
Fix bug 843531 - DPF: Access violation when DataSource is incorrectly constructed

MichaelNale:


src/ansys/dpf/core/data_sources.py
```diff
@@ -18,7 +18,8 @@
 )
 
 from ansys.dpf.core.check_version import version_requires
-
+from ansys.grpc.dpf import data_sources_pb2
```
PProfizi, line 21: @MichaelNale sadly if I remember correctly we on purpose do not directly import from `ansys.grpc.dpf`. However, the integration of it in `ansys-dpf-core`, I think this may not be an issue.

tests/test_datasources.py
```diff
@@ -61,6 +61,9 @@ def test_print_data_sources(allkindofcomplexity, server_type):
 
 
 def test_data_sources_from_data_sources(allkindofcomplexity, server_type):
+    with pytest.raises(ValueError) as e:
+        data_sources_false = dpf.core.DataSources(data_sources="Wrong Input", server=server_type)
```
PProfizi, line 65: @MichaelNale FYI, since I saw Codacy complaining, the pythonic thing to do here instead of having an unused variable is:
```suggestion
        _ = dpf.core.DataSources(data_sources="Wrong Input", server=server_type)
```

PR #1143, opened 2023-09-09
minimum dpf version is `0.4.0` and not 4.0

kmahajan-cadfem:
the version number noted in example was wrong
image


PProfizi:
Hi @kmahajan-cadfem,

First of all thnak you for this modification proposal as this is something we encourage whenever a user spots an error.
The version number stated is for the DPF server though, not the Python client. According to [this](https://dpf.docs.pyansys.com/version/stable/getting_started/compatibility.html#client-server), Ansys 2022 R2 does contain `DPF 4.0`, which requires `ansys-dpf-core>=0.5.0`.
Maybe the message should be improved though to also indicate the minimum `ansys-dpf-core` version.

PR #1146, opened 2023-09-11
Fix bug 843531 - DPF: Access violation when DataSource is incorrectly constructed

MichaelNale:


PProfizi (approved):

tests/test_datasources.py
```diff
@@ -61,6 +61,9 @@ def test_print_data_sources(allkindofcomplexity, server_type):
 
 
 def test_data_sources_from_data_sources(allkindofcomplexity, server_type):
+    with pytest.raises(ValueError) as e:
+        data_sources_false = dpf.core.DataSources(data_sources="Wrong Input", server=server_type)
+        assert "Data source must be gRPC data sources message type" in e
```
anslpa, line 66: Minor: if I were you, I would just assert that "gRPC" data sources" substring is contained in the error message.

src/ansys/dpf/core/data_sources.py
```diff
@@ -77,9 +78,12 @@ def __init__(self, result_path=None, data_sources=None, server=None):
                 self._internal_obj = core_api.data_processing_duplicate_object_reference(
                     data_sources
                 )
-            else:
+            elif hasattr(data_sources, "DESCRIPTOR") or isinstance(data_sources, int):
```
anslpa, line 81: You could also use if (hasattr(data_sources, "__module__") and data_sources.__module__ == 'data_sources_pb2')

anslpa: Note: int is for the case CLayer server, ctypes.c_void_p is returned (which is an int)

anslpa (approved):

PR #1164, opened 2023-09-18
Update the iso_contours example using iso_surfaces operator

MichaelNale:


examples/12-fluids/03-fluids_isosurface.py
```diff
@@ -67,11 +67,6 @@
 # We can finally use the mesh_cut operator on this specific variable.
 # We choose to cut the whole with 5 iso-surface equally spaced between min and max.
```
rafacanton, line 68: This needs update, as the use case has changed

examples/12-fluids/03-fluids_isosurface.py
```diff
@@ -87,18 +82,19 @@
...
+
+iso_surfaces = dpf.operators.mesh.iso_surfaces(field=P_S[0], mesh=whole_mesh, slice_surfaces=True,
+                                              vector_iso_values=vec_iso_surfaces).eval()
+
+for i in range(len(iso_surfaces)):
     P_S_step = dpf.Field(location=dpf.locations.overall, nature=dpf.common.natures.scalar)
-    P_S_step.append([min_pressure], i)
+    P_S_step.append(vec_iso_surfaces[i], i)
     P_S_step.name = "static pressure"
     P_S_step.unit = "Pa"
```
rafacanton, line 94: P_S_step doesn't need to be created by hand, as the second output of the operator should be giving it

examples/12-fluids/03-fluids_isosurface.py
```diff
@@ -85,16 +85,16 @@
...
-for i in range(len(iso_surfaces)):
-    P_S_step = dpf.Field(location=dpf.locations.overall, nature=dpf.common.natures.scalar)
-    P_S_step.append(vec_iso_values[i], i)
-    P_S_step.name = "static pressure"
-    P_S_step.unit = "Pa"
+iso_surfaces_meshes = dpf.operators.mesh.iso_surfaces(field=P_S[0], mesh=whole_mesh, slice_surfaces=True,
+                                              vector_iso_values=vec_iso_values).outputs.meshes()
+
+iso_surfaces_fields = dpf.operators.mesh.iso_surfaces(field=P_S[0], mesh=whole_mesh, slice_surfaces=True,
+                                              vector_iso_values=vec_iso_values).outputs.fields_container()
```
rafacanton, line 92: With this you are evaluating the operator again, which can be very costly for large models (we shouldn't put it in an example). Instead, we should instantiate the operator once and then get the two outputs

rafacanton (approved):

PR #1166, opened 2023-09-19
Produce OS-specific wheels 

### LINKED ISSUE #1165: Create OS-specific wheels as pipeline artifacts
Following the integration of `ansys-dpf-gatebin` binaries in `ansys-dpf-core`, it does create some issues in some deployment contexts, requiring to create OS-specific `ansys_dpf_core` wheels:
 
- `ansys_dpf_core-X.Y.Z-py3-none-win_amd64.whl`, contains gatebin for windows
- `ansys_dpf_core-X.Y.Z-py3-none-manylinux_2_17_x86_64.whl`, contains gatebin for manylinux_2_17
- `ansys_dpf_core-X.Y.Z-py3-none-manylinux1_x86_64.whl`, contains gatebin for manylinux1
- `ansys_dpf_core-X.Y.Z-py3-none-any.whl` (for Darwin a.k.a. macOS), does not contain gatebin

The pipelines thus need changing so that these are output artifacts, most likely of the **tests** jobs, during the action which creates the wheel:

- [ ] Jobs run on a Linux runner will output `manylinux1`, `manylinux_2_17`, and `any` versions of the wheel as artifacts
- [ ] Jobs run on a Windows runner will output `win` versions of the wheel as artifacts

The four wheels should also:

- [ ] be released to PyPI during the release pipeline
- [ ] put as assets of the GH Release.

PProfizi:
Closes #1165 

Adds a .`ci/build_wheel.py` Python script which takes a platform argument (as well as a whether to create a wheelhouse)
It is used in separate jobs to create and output the `any` and `manylinux1` wheels.
It is used everywhere else to build, test and output the `win_amd64` wheel for Windows runners, and the `manylinux_2_17` wheel for Linux runners.
The `any` wheel is also tested.


cbellot000 (commented):
@PProfizi could you please give us an overview of what tests/examples are ran on which wheel?
I've not seen any code change to make the any wheel work, were there already ifs in the code to find the binaries in the installer when the pydpf-coe wheel doesn't have any binary?

PProfizi:
> @PProfizi could you please give us an overview of what tests/examples are ran on which wheel? I've not seen any code change to make the any wheel work, were there already ifs in the code to find the binaries in the installer when the pydpf-coe wheel doesn't have any binary?

@cbellot000 I've added a tests job for the `any` wheel on Linux and on Windows, and the logic when not finding gatebin binaries is [here](https://github.com/ansys/pydpf-core/blob/2cd30046d6ddec7f46a963f6d41b7601016b7f07/src/ansys/dpf/gate/load_api.py#L98). There is an error message which might require an update but otherwise it should go look into the installation yes.

.ci/build_wheel.py
```diff
@@ -0,0 +1,94 @@
...
+    binaries_to_move = []
+    moved = []
+    if "win" in requested_platform or "any" == requested_platform:
+        # Move linux binaries
+        binaries_to_move.extend(["libAns.Dpf.GrpcClient.so", "libDPFClientAPI.so"])
+    if "linux" in requested_platform or "any" == requested_platform:
+        # Move windows binaries
+        binaries_to_move.extend(["Ans.Dpf.GrpcClient.dll", "DPFClientAPI.dll"])
+    if "any" == requested_platform:
+        binaries_to_move.extend(["_version.py"])
```
anslpa, line 60: @PProfizi _version.py is not needed for the other packages ?

.github/workflows/tests.yml
```diff
@@ -7,7 +7,7 @@ on:
       python_versions:
         required: false
         type: string
-        default: '["3.7"]'
+        default: '["3.8"]'
```
anslpa, line 10: @PProfizi Minor: just that you know, I've looked for "3.7" references into the repository and it looks we still have a few Python 3.7 references

image

.github/workflows/pydpf-post.yml
```diff
@@ -73,18 +73,48 @@ jobs:
         with:
           python-version: "3.8"
 
-      - name: "Build Core Package"
-        uses: ansys/pydpf-actions/build_package@v2.3
+      - name: "Install ansys-dpf-core build requirements"
+        run: pip install -r requirements/requirements_build.txt
+
+      - name: "Build ansys-dpf-core wheel"
```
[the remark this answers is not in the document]
PProfizi: @anslpa sure, but this is refactoring and not the priority for now. In time yes we can make an action out of this, which will have to reside in pydpf-actions but at the same time it is an action only for pydpf-core... I need to investigate the best way to simplify this.
We cannot make a callable `.yml` directly in the repo either as GH only allows for two levels which we already have.

.ci/build_wheel.py
PProfizi: @anslpa to make the `any` wheel, I need to completely remove the `gatebin` folder, including this file.
The tests do pass when using this `any` wheel.
I even think we should remove those `_version.py` files from `ansys.grpc.dpf`, `ansys.dpf.gatebin` and `ansys.dpf.gate` altogether.

.github/workflows/tests.yml
PProfizi: @anslpa the problem I have with removing these options from the dropdown lists is that although we do not support 3.7 anymore for our latest packages versions, people may still want to report issues or questions about an older package which still supported it.

.ci/build_wheel.py
```diff
@@ -0,0 +1,94 @@
+# This script generates the different versions of the ansys-dpf-core wheels based on a given input.
+# Input can be one of ["any", "win", "manylinux1", "manylinux_2_17"]
+
+import argparse
```
cbellot000, line 4: if I remember correctly, wheels used to be build by dpf-action repo, what is done where now?

cbellot000: what about pydpf-post?

.github/workflows/gate.yml
```diff
@@ -71,23 +71,85 @@ jobs:
...
+      - name: "Setup Python"
         uses: actions/setup-python@v4.6.0
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: "Build Package"
-        uses: ansys/pydpf-actions/build_package@v2.3
+      - name: "Install requirements"
+        run: pip install -r requirements/requirements_build.txt
+
```
cbellot000, line 81: what is the purpose of the gate.yml workflow now ?

.github/workflows/pydpf-post.yml
```diff
@@ -73,18 +73,48 @@ jobs:
...
+        shell: bash
+        run: |
+          if [ ${{ matrix.os }} == "ubuntu-latest" ]; then
```
cbellot000, line 82: is there any place where we have all wheels available (link in pypi) and we let pip pick the right one?

.ci/build_wheel.py
PProfizi: > if I remember correctly, wheels used to be build by dpf-action repo, what is done where now?

@cbellot000 they indeed used to be created by an action in the dpf-action repo. Now I've changed to have the logic directly in the pydpf-core yml files. I will consider making a common action later but since this is particular to pydpf-core, it would not have a big advantage to move it to the pydpf-actions

PProfizi: > what about pydpf-post?

Right now the pydpf-post regular CI just clones the pydpf-core repo and installs it as-is since the dependency is on the git repository directly. I'll have to make a PR on PyDPF-Post to properly build the ansys-dpf-core wheel based on the context, but it can be done after this.
The pydpf-post Release CI gets an ansys-dpf-core wheel from pip, so it requires no change.

.github/workflows/gate.yml
PProfizi: > what is the purpose of the gate.yml workflow now ?

@cbellot000  ah yes I'll remove it

.github/workflows/pydpf-post.yml
PProfizi: @cbellot000 this is not tested in the CI, no.
We know that pip will pick from most-specific to more-generic, I tested that for other libraries which also do things this way.

.ci/build_wheel.py
cbellot000: Okay, will there be any time with a broken pipeline in pydpf-post? (it's just to make sure we capture evrything)

Also if this will be used in pydpf-post, I guess you'll move it to pydpf-actions after?

.github/workflows/pydpf-post.yml
cbellot000: maybe we could at least run the example like that, then (at least in the release pipeline)? if not, we are not really testing a user environment

.ci/build_wheel.py
PProfizi: > Okay, will there be any time with a broken pipeline in pydpf-post? (it's just to make sure we capture evrything)
> 
> Also if this will be used in pydpf-post, I guess you'll move it to pydpf-actions after?

Hi @cbellot000, sorry I do not understand the first sentence.

As for pydpf-post, indeed once I have a build_core action I can move it to the pydpf-actions, but this is really a consolidation step which is not really required at this point.

PProfizi: See [here](https://github.com/ansys/pydpf-post/actions/runs/6479253578) for a run of the PyDPF-Post CI while targeting this branch specifically in the requirements, meaning this branch will be cloned when pip installing dependencies of `ansys-dpf-post`.

cbellot000 (approved):

PR #1203, opened 2023-10-12
Revert "Update the iso_contours example using iso_surfaces operator"

MichaelNale:
Reverts ansys/pydpf-core#1164

### DIFF
--- a/examples/12-fluids/03-fluids_isosurface.py
+++ b/examples/12-fluids/03-fluids_isosurface.py
@@ -64,9 +64,13 @@ pl.show_figure(cpos=cpos_mesh_variable, show_axes=True)
 ###############################################################################
 # Evaluate iso-surfaces
 # ~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


PProfizi:
Necessary due to incompatibility between this new version of the example and DPF 2024.1.pre0.
This PR should be revert itself once ansys-dpf-core for DPF 2024.1.pre1 is under preparation.

rafacanton:
@PProfizi I don't understand. Wouldn't it be better to just mark it as limited to versions >=7.1?

PProfizi:
@rafacanton sadly no, since it already appears in the doc in its older version, we can't just bump the required version to a DPF verion that is not released yet. The point is we don't want to it disappear upon the release of a new ansys-dpf-core, which is what would happen if we mark it as requiring a higher DPF version than what is currently available.

PProfizi:
When generating the doc, we filter out the examples that require a DPF version superior to what is being used to generate the doc.

PR #1205, opened 2023-10-12
Refactor version handling

PProfizi:
This PR proposes a refactor of `ansys.dpf.core._version.py`'s `server_to_ansys_version` to enable handling `XXXX.Y.Z` DPF versioning while retaining retro-compatibility.
It also removes `server_to_ansys_grpc_dpf_version` as `ansys.grpc.dpf` is now shipped within `ansys-dpf-core`.
Instead, what is checked when connecting to a server via gRPC is whether the DPF server we try to connect to has a version at least equal to the `min_server_version` variable, which is the minimal DPF version supported by the current `ansys-dpf-core` package.

### DIFF
--- a/src/ansys/dpf/core/_version.py
+++ b/src/ansys/dpf/core/_version.py
@@ -8,16 +8,6 @@ __version__ = ".".join(map(str, version_info))
 # Minimal DPF server version supported
 min_server_version = "4.0"
 
-server_to_ansys_grpc_dpf_version = {
-    "1.0": "==0.2.2",
-    "2.0": "==0.3.0",
-    "3.0": ">=0.4.0",
-    "4.0": ">=0.5.0",
-    "5.0": ">=0.6.0",
-    "6.0": ">=0.7.0",
-    "7.0": ">=0.8dev",
-}
-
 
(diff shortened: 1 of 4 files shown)


src/ansys/dpf/core/_version.py
```diff
@@ -4,27 +4,31 @@
...
-    "5.0": "2023R1",
-    "6.0": "2023R2",
-    "6.1": "2023R2",
-    "6.2": "2023R2",
-    "7.0": "2024R1",
-    "7.1": "2024R1",
-}
+
+class ServerToAnsysVersion:
+    legacy_version_map = {
```
cbellot000, line 13: why is that called legacy?

cbellot000 (approved):

src/ansys/dpf/core/_version.py
PProfizi: @cbellot000 the legacy versioning logic, vs the new one following the Ansys product versioning, but I can change the name, no problem

PProfizi: I think calling it legacy makes it clear it should not be changed.

PR #1243, opened 2023-10-26
Don't create LocalField when using InProcess

ansys-akarcher:
The Local field primary goal is to copy the data when a field is remote, and very costly to access, which is not the case when the DPF server is used InProcess.

For now I didn't find an impactful and minimal way to get faster on get_entity_data_* methods.


PProfizi (approved):

src/ansys/dpf/core/field_base.py
```diff
@@ -494,6 +494,12 @@ def _get_data(self, np_array=True):
     def _set_data(self, data):
         pass
 
+    def __enter__(self):
```
cbellot000, line 497: @ansys-akarcher is that needed to do with statement?

cbellot000 (approved):

src/ansys/dpf/core/field_base.py
ansys-akarcher: @cbellot000 
__enter__ and __exit__ are indeed required as now as_local_field() may be returning either a Field Object or a LocalField (which was defining __enter__ and __exit__)

PR #1252, opened 2023-11-06
Add CustomTypeFieldsContainer

PProfizi:
The `CustomTypeFieldsContainer` is an equivalent of the `FieldsContainer` but for a collection of `CustomTypeField` instead of `Field.`
The `CustomTypeField` is a equivalent of the `Field` with the option to select a data type:
- A `CustomTypeField` of `dtype=np.float` can be an equivalent to the `Field`.
- A `CustomTypeField` of `dtype=np.int` can be an equivalent to the `PropertyField`.

NB:
- the `Collection` API does not currently accept a `CustomTypeFieldsContainer` as input, making all the usual setters for a FieldsContainer fail when used here. We are now planning to restrict this PR to the exposition as a potential input/output of an operator.
- the server description returned, used for the `str` method, is empty.


src/ansys/dpf/core/custom_type_fields_container.py
```diff
@@ -0,0 +1,589 @@
...
+
+    def __time_complex_label_space__(self, timeid: int = None, complexid: int = None) -> dict:
+        label_space = {}
+        if timeid is not None:
+            label_space["time"] = timeid
+        if complexid is not None:
+            label_space["complex"] = complexid
+        return label_space
+
+    def get_fields(self, label_space: dict[str, int]) -> list[CustomTypeField]:
```
rafacanton, line 134: I guess that it is intentional, but wouldn't it make more sense to specialize the apis? I mean, `get_customfields` instead of `get_fields`, and the same for the rest

tests/test_custom_type_fields_container.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
+
+import conftest
+
+import ansys.dpf.core as dpf
+
```
rafacanton, line 6: Can we have a test of getting the CustomTypeFC as output from an operator?

src/ansys/dpf/core/dpf_operator.py
```diff
@@ -325,6 +326,11 @@ def _type_to_output_method(self):
                 self._api.operator_getoutput_custom_type_field,
                 "field",
             ),
+            (
+                custom_type_fields_container.CustomTypeFieldsContainer,
+                self._api.operator_getoutput_custom_type_fields_container,
+                "fields_container",
```
rafacanton, line 332: "custom_type_fields_container"?

tests/test_custom_type_fields_container.py
PProfizi: @rafacanton, yes, it will be. I need the `Compression` plugin to be updated for that end though.

src/ansys/dpf/core/custom_type_fields_container.py
PProfizi: @rafacanton yes, it was intentional on my part to have its usage as similar as possible to that of a `FieldsContainer` as to me they are two different implementations of the same thing, so they should have as much a similar API as possible.
I had to remove those methods though as server-side none of the `collection_capi` methods understand a CustomTypeFieldsContainer.

rafacanton (approved):

PR #1275, opened 2023-11-14
Always connect to a server via LegacyGrpc when on macOS

### LINKED ISSUE #1274: Automatically choose LegacyGrpc when on macOS

When connecting to a remote server via `connect_to_server`, test if local machine is macOS and if so, set LegacyGrpc by default. 
Apparently trying out CGrpc first and setting back to LegacyGrpc after failure is not working properly all the time. 

_No response_

_No response_

PProfizi:
Closes #1274

### DIFF
--- a/src/ansys/dpf/core/server.py
+++ b/src/ansys/dpf/core/server.py
@@ -7,6 +7,7 @@ import os
 import socket
 import weakref
 import copy
+import platform
 import inspect
 import warnings
 import traceback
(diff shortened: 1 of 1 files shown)


src/ansys/dpf/core/server.py
```diff
@@ -355,6 +359,10 @@ def connect():
         dpf.core._server_instances.append(weakref.ref(server))
         return server
 
+    # Enforce LegacyGrpc when on macOS
+    if platform.system() == 'Darwin':
+        config = dpf.core.AvailableServerConfigs.LegacyGrpcServer
```
rafacanton, line 364: @PProfizi maybe we could add mac agents to the CI/CD (at least for the current server version) to be able to test all this. I don't know how difficult would this be, as all the current testing matrices that we have tend to run things for CLayer InProcess, CLayer gRPC and LegacyGrpc, but it could be worth doing if we start t make these kind of distinctions

rafacanton (approved):

PR #1276, opened 2023-11-15
Refactor misc.py

### LINKED ISSUE #1273: Check AWP_ROOT path for DPF install

Check AWP_ROOT path for DPF install

_No response_

_No response_

PProfizi:
Refactor misc.py/_find_latest_ansys_versions as the same logic is now at the gate level in load_api.py.

Originally showed the required changes on pygate to close #1273


src/ansys/dpf/gate/load_api.py
```diff
@@ -51,38 +51,51 @@ def _pythonize_awp_version(version):
...
-        if not awp_version.isnumeric():
-            continue
-        ansys_path = os.environ.get("AWP_ROOT" + awp_version)
-        if ansys_path:
-            installed_packages_list[
-                packaging.version.parse(_pythonize_awp_version(awp_version))
-            ] = ansys_path
+    path_per_version = _paths_to_dpf_in_unified_installs(path_per_version)
+
+    path_per_version = _paths_to_dpf_server_library_installs(path_per_version)
```
rafacanton, line 59: @PProfizi As we are adding entries to the same dictionary (`path_per_version`), do we risk overwritten already existing entries? Or this is the expected behavior as per your proposal?

src/ansys/dpf/gate/load_api.py
```diff
@@ -51,38 +51,51 @@ def _pythonize_awp_version(version):
...
+
+    if len(path_per_version) > 0:
```
rafacanton, line 61: if not, raise an error? Server not found or something similar?

src/ansys/dpf/gate/load_api.py
```diff
@@ -51,38 +51,51 @@ def _pythonize_awp_version(version):
...
+    for awp_version in awp_versions:
+        if not awp_version.isnumeric():
+            continue
+        ansys_path = os.environ.get("AWP_ROOT" + awp_version)
+        if ansys_path:
+            # Check that this ansys path exists
+            if not os.path.isdir(ansys_path):
+                continue
+            # Check that it contains a DPF install
+            if not os.path.exists(os.path.join(ansys_path, _get_path_in_install())):
```
rafacanton, line 93: `_get_path_in_install` should be enriched to also check the existence of the `dpf` folder, where the majority of libraries are now placed

src/ansys/dpf/gate/load_api.py
PProfizi: @rafacanton `AWP_ROOTXXX` variables give 'production' versions (such as `2024.2`), see `_pythonize_awp_version`, while `ansys_dpf_server` versions are always pre-production (such as `2024.2.pre0`) so IMO they should not intersect...

src/ansys/dpf/gate/load_api.py
PProfizi: @rafacanton this should return nothing, which is a case treated by `ansys.dpf.core.misc.find_ansys()`

src/ansys/dpf/gate/load_api.py
PProfizi: @rafacanton we could, yet checks are usually made in methods which call `_get_path_in_install`:
- `gate.load_api._get_api_path_from_installer_or_package`, 
- `core.server_types._run_launch_server_process`, 
- `core.server_types._verify_ansys_path_is_valid`, 
- `core.server_types._get_dll_path`, 
- `core.server_factory.ServerFactory.get_server_type_from_config`,

That would mean refactoring all of those.
Still I agree that the current checks may not be enough. We should check that a given list of DLLs (TBD) is actually present.
This can be added to the objectives of this issue/PR, what do you think?

PProfizi: @rafacanton while implementing a test for `dpf/bin/winx64` or `dpf/bin/linx64` folder, I realized that we may have an issue in `ServerFactory.get_server_type_from_config` which basically enforces use of `AWP_ROOT242` when in non-legacy Grpc to define the aisol path to add to PATH... I think this is quite a problem. I also noticed that we still use `_find_outdated_ansys_version()` which tests for Ansys versions below 221... I do not think we still need that as it was used to force `LegacyGrpc` connections for servers 221 or below while now we could argue that the client itself does not support servers below 222. I think there may be a lot of refactoring to do here.

PProfizi: Note: adding the test for a dpf folder, yet **it appears consistently only starting with 231**.

docs/source/getting_started/index.rst
```diff
@@ -66,7 +66,7 @@ Install DPF Server
 
 For more information about the license terms, see :ref:`DPF Preview License Agreement<target_to_license_terms>`.
 
-To use a remote license, change or set the ``ANSYSLMD_LICENSE_FILE`` environment
+To use a remote license set the ``ANSYSLMD_LICENSE_FILE`` environment
```
PProfizi, line 69:
```suggestion
To use a remote license, change or set the ``ANSYSLMD_LICENSE_FILE`` environment
```

src/ansys/dpf/core/misc.py
```diff
@@ -107,35 +107,7 @@ def _pythonize_awp_version(version):
...
-                installed_packages_list[
-                    packaging.version.parse(module.__version__)
-                ] = module.__path__[0]
-            except ModuleNotFoundError:
-                pass
-            except AttributeError:
-                pass
-    if len(installed_packages_list) > 0:
-        return installed_packages_list[sorted(installed_packages_list)[-1]]
+    return load_api._find_latest_ansys_versions()
```
rafacanton, line 110: @PProfizi The only final change is then calling the gate method from `misc`, correct?

PProfizi: @rafacanton yes that is the only thing left in this PR, which is thus independent and can be merged as-is.

rafacanton (approved):

PR #1277, opened 2023-11-15
Detect binout files automatically

PProfizi:
Part of work for ansys/pydpf-post#534.

Improves previous PR for automatic result key for known files without extensions (d3plot and binout)

### DIFF
--- a/src/ansys/dpf/core/data_sources.py
+++ b/src/ansys/dpf/core/data_sources.py
@@ -116,10 +116,16 @@ class DataSources:
         ['/tmp/file.rst']
 
         """
-        # Handle the case of files without extension, such as the LS-DYNA d3plot file
+        # Handle the case of files without extension, such as the LS-DYNA d3plot and binout files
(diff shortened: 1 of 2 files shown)


src/ansys/dpf/core/data_sources.py
```diff
@@ -116,15 +116,26 @@ def set_result_file_path(self, filepath, key=""):
...
+    @staticmethod
+    def guess_result_key(filepath: str) -> str:
+        """Guess result key for files without a file extension."""
+        d3plot_key = "d3plot"
+        binout_key = "binout"
+        # Handle files without extension
+        if d3plot_key in os.path.basename(filepath):
+            return d3plot_key
+        elif binout_key in os.path.basename(filepath):
+            return binout_key
```
rafacanton, line 136: Maybe we could generalize this in a list ["d3plot", "binout"], and loop over it, so it can be easily extensible

rafacanton (approved):

PR #1284, opened 2023-11-21
Optimize CGrpcServer getters

PProfizi:
Edit: the actual issue was found to be too many gRPC calls to ResultInfo getters when using a `LegacyGrpcServer`. The solution found was to add a cache decorator to functions called repeatedly.
We thus went down from 1004 gRPC calls waiting a total of `104s`, to 166 gRPC calls waiting a total of `15s`.

By looking at the `GrpcServer` calls, caching the `os` and `version` properties also decreases total time from `50s` to `15s`.

There may be further improvements to be made that I did not see with the simple `print(model)` command.

NB: I did enforce a limit of 50 for the cache while testing, this is TBD
NB2: the first initiative consisting of improving the description given by the server could still be beneficial.
NB3: all changes related to gate have been made in a different PR server-side

Old:
Summary:

Upon arrival of Fluid results, the `ResultInfo.__str__` method was enhanced to also show the locations of each available result as they are dynamic, as well as create a map of qualifier names vs IDs, as well as properly detect the fluid physics type.
This creates a performance deficit of 1200% when in LegacyGrpc (without any actual network delay) due to the number of requests made just when requesting for available results.
The goal of this PR is to revert to the server description of the `ResultInfo` when in LegacyGrpc to bypass this performance bottleneck, while server-side changes are to be made to:
- add result location to available result descriptions
- add a map of available qualifiers
- correctly show the physics type when fluid
- switch to the server description of the `ResultInfo` in all cases, potentially with some very minor formatting

### DIFF
--- a/src/ansys/dpf/gate/result_info_grpcapi.py
+++ b/src/ansys/dpf/gate/result_info_grpcapi.py
@@ -26,6 +26,7 @@ class ResultInfoGRPCAPI(result_info_abstract_api.ResultInfoAbstractAPI):
         result_info._deleter_func = (_get_stub(result_info._server).Delete, lambda obj: obj._internal_obj)
 
     @staticmethod
+    @functools.lru_cache(maxsize=50, typed=False)
     def list(result_info):
         from types import SimpleNamespace
         server = result_info._server
--- a/src/ansys/dpf/gate/result_info_grpcapi.py
+++ b/src/ansys/dpf/gate/result_info_grpcapi.py
@@ -222,6 +222,7 @@ class ResultInfoGRPCAPI(result_info_abstract_api.ResultInfoAbstractAPI):
         return ResultInfoGRPCAPI.list(result_info).main_title
 
     @staticmethod
+    @functools.lru_cache(maxsize=50, typed=False)
     def result_info_get_string_property(result_info, property_name):
         from ansys.grpc.dpf import result_info_pb2
         request = result_info_pb2.GetStringPropertiesRequest()
--- a/src/ansys/dpf/gate/result_info_grpcapi.py
+++ b/src/ansys/dpf/gate/result_info_grpcapi.py
@@ -1,3 +1,5 @@
+import functools
+
 from ansys.dpf.gate.generated import result_info_abstract_api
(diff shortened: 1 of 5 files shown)


src/ansys/dpf/gate/result_info_grpcapi.py
```diff
@@ -1,3 +1,5 @@
+import functools
+
```
PProfizi, line 2:
```suggestion
```

src/ansys/dpf/gate/result_info_grpcapi.py
```diff
@@ -24,6 +26,7 @@ def init_result_info_environment(result_info):
         result_info._deleter_func = (_get_stub(result_info._server).Delete, lambda obj: obj._internal_obj)
 
     @staticmethod
+    @functools.lru_cache(maxsize=50, typed=False)
```
PProfizi, line 29:
```suggestion
```

src/ansys/dpf/gate/result_info_grpcapi.py
```diff
@@ -55,6 +58,7 @@ def list(result_info):
         return response
 
     @staticmethod
+    @functools.lru_cache(maxsize=50, typed=False)
```
PProfizi, line 61:
```suggestion
```

src/ansys/dpf/gate/result_info_grpcapi.py
```diff
@@ -219,6 +223,7 @@ def result_info_get_main_title(result_info):
         return ResultInfoGRPCAPI.list(result_info).main_title
 
     @staticmethod
+    @functools.lru_cache(maxsize=50, typed=False)
```
PProfizi, line 226:
```suggestion
```

src/ansys/dpf/gate/result_info_grpcapi.py
```diff
@@ -227,6 +232,7 @@ def result_info_get_string_property(result_info, property_name):
         return _get_stub(result_info._server).GetStringProperties(request).properties[property_name]
 
     @staticmethod
+    @functools.lru_cache(maxsize=50, typed=False)
```
PProfizi, line 235:
```suggestion
```

cbellot000 (approved):

PR #1296, opened 2023-11-28
Improve Workflow.connect_with docstring

PProfizi:
The previous docstring was wrong. Using this opportunity to clarify the behavior and add an example in the docstring.

### DIFF
--- a/src/ansys/dpf/core/workflow.py
+++ b/src/ansys/dpf/core/workflow.py
@@ -630,19 +630,24 @@ class Workflow:
 
     @version_requires("3.0")
     def connect_with(self, left_workflow, output_input_names=None):
-        """Chain 2 workflows together so that they become one workflow.
+        """Prepend a given workflow to the current workflow.
 
-        The one workflow contains all the operators, inputs, and outputs
-        exposed in both workflows.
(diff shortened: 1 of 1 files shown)


FedericoNegri (commented):
Thanks for updating the doc, looks good to me.

src/ansys/dpf/core/workflow.py
```diff
@@ -630,19 +630,24 @@ def output_names(self):
...
         left_workflow : core.Workflow
-            Second workflow's outputs to chained with this workflow's inputs.
+            The given workflow's outputs are chained with the current workflow's inputs.
         output_input_names : str tuple, str dict optional
-            Input name of the left_workflow to be cained with the output name of this workflow.
-            The default is ``None``, in which case the inputs in the left_workflow with the same
-            names as the outputs of this workflow are chained.
+            Map used to connect the outputs of the given workflow to the inputs of the current
+            workflow.
+            Check the available inputs and outputs names of each workflow using
```
JennaPaikowsky, line 647:
```suggestion
            Check the names of available inputs and outputs for each workflow using
```

JennaPaikowsky (approved):

PR #1300, opened 2023-11-29
Display Workflows and save as file

PProfizi:
Proposal for a `Workflow` viewer in PyDPF.
This uses `graphviz` rendering and the native system image viewer to show the image obtained with `graphviz`.

To be discussed.


rafacanton:
Generic comment: although I have the graphviz Python dependency in the venv, this was not enough to render the image. It raised this error: `graphviz.backend.execute.ExecutableNotFound: failed to execute WindowsPath('dot'), make sure the Graphviz executables are on your systems' PATH`, so I needed to manually install graphviz from [here](https://graphviz.org/download/). If we want this out the box, maybe we need an additional dependency that installs graphviz server.

src/ansys/dpf/core/workflow.py
```diff
@@ -806,6 +809,53 @@ def create_on_other_server(self, *args, **kwargs):
...
+            If None, the system's default image viewer is used.
+
+        title:
+            Name to use in intermediate files and in the viewer.
+        """
+        if title is None:
+            name = f"workflow_{repr(self).split()[-1][:-1]}"
+        else:
+            name = title
+        file_path = os.path.join(os.getcwd(), f"{name}.dot")
```
rafacanton, line 832: Do we need a file? Maybe we could add an api similar to `self._api.work_flow_export_graphviz` that generates a big string (the contents of the dot file) that are then passed to graphviz without files

src/ansys/dpf/core/workflow.py
```diff
@@ -806,6 +809,53 @@ def create_on_other_server(self, *args, **kwargs):
...
+        try:
+            import graphviz
+        except ImportError:
+            raise ValueError(f"To render workflows using graphviz, run 'pip install graphviz'.")
+        graphviz.render(engine='dot', format='png', filepath=file_path)
+        file_path = file_path+".png"
+
+        # View workflow
+        if viewer is None:
+            graphviz.view(filepath=file_path)
```
rafacanton, line 845: same here

rafacanton:
This is what I get with (viewer="paraview"):
image

PProfizi:
@FedericoNegri

PProfizi:
> This is what I get with (viewer="paraview"): image

I removed the PyVista viewer as it had no added value and just made everything more complicated.

src/ansys/dpf/core/workflow.py
PProfizi: @rafacanton I agree, we would need an operator which just returns the contents of the DOT file. But that would mean a new operator, and managing how to feed this big string to the renderer, which currently only accepts files, so we would have to mimick a file... 
I think this is for a future PR when we see the need.
Right now I improved the behavior so we clean-up the DOT file by default.

PProfizi:
> Generic comment: although I have the graphviz Python dependency in the venv, this was not enough to render the image. It raised this error: `graphviz.backend.execute.ExecutableNotFound: failed to execute WindowsPath('dot'), make sure the Graphviz executables are on your systems' PATH`, so I needed to manually install graphviz from [here](https://graphviz.org/download/). If we want this out the box, maybe we need an additional dependency that installs graphviz server.

Added the requirement to the dosctring

rafacanton (approved):

PR #1314, opened 2023-12-08
improve runtime client config

cbellot000:
BUG: The runtime client config is only client side: it should not start a server
when a client is created a after client setting are changed: the client should take into account these changes


src/ansys/dpf/core/server_types.py
```diff
@@ -1086,7 +1100,7 @@ def get_api_for_type(self, capi, grpcapi):
         return grpcapi
 
     def create_stub_if_necessary(self, stub_name, stub_type):
-        if not (stub_name in self._stubs.keys()):
+        if self.channel and not (stub_name in self._stubs.keys()):
```
cbellot000, line 1103:
```suggestion
        if self.channel and not stub_name in self._stubs:
```

src/ansys/dpf/core/settings.py
```diff
@@ -121,8 +125,24 @@ def get_runtime_client_config(server=None):
         with Ans.Dpf.GrpcClient configuration.
 
     """
-    base = core.BaseService(server, load_operators=False)
-    return base.get_runtime_client_config()
+    from ansys.dpf.core.runtime_config import RuntimeClientConfig
+    from ansys.dpf import core
```
cbellot000, line 129:
```suggestion
    from ansys.dpf import core as root
```

src/ansys/dpf/core/settings.py
```diff
@@ -121,8 +125,24 @@ def get_runtime_client_config(server=None):
...
+    if server is None:
+        server = core.SERVER
```
cbellot000, line 131:
```suggestion
        server = root.SERVER
```

src/ansys/dpf/gate/misc.py
```diff
@@ -1,11 +1,33 @@
 DEFAULT_FILE_CHUNK_SIZE = None
 COMMON_PROGRESS_BAR = None
-_CLIENT_CONFIG = {}
+
```
PProfizi, line 3: @cbellot000 do we have an equivalent PR on pygate?

cbellot000: yes

src/ansys/dpf/core/core.py
```diff
@@ -504,21 +504,6 @@ def release_dpf(self):
         else:
             error = self._api.data_processing_release(1)
 
-    def get_runtime_client_config(self):
```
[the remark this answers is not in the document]
cbellot000: yes this baseService is an internal class, users shouldn't use it

rafacanton (approved):

PR #1335, opened 2023-12-22
Expose Collection.name

PProfizi:
Allows to get and set names for Collections (missing api for LegacyGrpc)
Necessary to allow exporting named custom results to VTU.

TODO:
- [x] Add gRPC


cbellot000:
@PProfizi is it straightforward to add it in legacy grpc as well? We now have big customers using it, if we could prevent having too  many differences between our server config I think it would be good

PProfizi:
> @PProfizi is it straightforward to add it in legacy grpc as well? We now have big customers using it, if we could prevent having too many differences between our server config I think it would be good

Hey @cbellot000, the goal is to implement it for all types on connections, hence the testing being activated for all 3 types. Now it works for InProcess, my next step will be to implement changes so it is available via gRPC too.

src/ansys/dpf/gate/collection_grpcapi.py
```diff
@@ -95,6 +95,18 @@ def collection_get_num_labels(collection):
...
+        return CollectionGRPCAPI._list(collection).name
+
+    @staticmethod
+    def collection_set_name(collection, name):
+        from ansys.grpc.dpf import collection_pb2
+        request = collection_pb2.UpdateCollectionRequest()
+        request.collection.CopyFrom(collection._internal_obj)
+        request.string_properties.update({"name": name})
+        _get_stub(collection._server).Update(request)
+
```
PProfizi, line 109:
```suggestion
```

cbellot000 (approved):

PR #1339, opened 2024-01-05
Improve data tree

cbellot000:


tests/test_data_tree.py
```diff
@@ -385,3 +385,33 @@ def test_attribute_errors_data_tree(server_type):
...
+        to_fill.list_double = [1.5, 2.5]
+        to_fill.list_string = ["hello", "bye"]
+    assert data_tree.get_as("list_string") == "hello;bye"
+    assert data_tree.get_as("int", int) == 1
+    assert data_tree.get_as("int", bool) == True
+    assert data_tree.get_as("double", float) == 1.0
+    assert data_tree.get_as("string", str) == "hello"
+    assert data_tree.get_as("list_int", dpf.types.vec_int) == [1, 2]
+    assert data_tree.get_as("list_double", dpf.types.vec_double) == [1.5, 2.5]
+    assert data_tree.get_as("list_string", dpf.types.vec_string) == ["hello", "bye"]
```
PProfizi, line 417: @cbellot000 quick question: why is this not the default output type for a list of strings?

cbellot000: The data tree doesn't hold typed data, it only holds string, so we can only return strings by default

PProfizi (approved):

PR #1341, opened 2024-01-05
feat: add Result.on_region_scoping

PProfizi:
Edit: This PR adds the `Result.on_region_scoping` method in an effort to improve the result related API.

This PR aims at improving the Core APIs regarding working with regions (zones, parts...).

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Use Case | using core   operators | using core   helpers | using Post   API
-- | -- | -- | --
Query the   mesh for a given fluid zone | dpf.operators.mesh.mesh_provider(       data_sources=model,       region_scoping=single_zone_list,   ).eval()   --> DOES NOT ACCEPT A LIST OF ZONES | provider =   model.metadata.mesh_provider   provider.inputs.region_scoping.connect(single_zone_list)   mesh =   provider.eval() --> NOT A PROPER HELPER | Provide   mesh query API -> add example
Query a   result on a given fluid zone | dpf.operators.result.static_pressure(       data_sources=model,       region_scoping=zone_list,   ).eval() | model.results.static_pressure.on_region_scoping([4]).eval() | simulation.static_pressure(zone_ids=zone_list)
Query the   mesh for a given lsdyna part | dpf.operators.mesh.mesh_provider(       data_sources=model,       region_scoping=single_part_list,   ).eval()   --> ACCEPTS A LIST OF PARTS | provider =   model.metadata.mesh_provider   provider.inputs.region_scoping.connect(single_part_list)   mesh =   provider.eval() --> NOT A PROPER HELPER | Provide   mesh query API -> add example
Query a   result for a given lsdyna part | op =   dpf.operators.result.displacement(       data_sources=model,   )   op.connect(25,   part_list)   fc =   op.eval()   -->   non-fluid result operators do not have a region_scoping pin exposed |   |  

</div>

</body>

</html>

Current observations:

1. behavior of operator meshes_provider is not the same between fluid zones and lsdyna parts: for lsdyna the mesh is not labelled per part
3. behavior of operator mesh_provider is not the same between fluid zones and lsdyna parts: it accepts a list of parts for lsdyna and returns the merged mesh; it throws for multiple fluid zones
4. the region_scoping pin is not exposed for non-fluid operators, meaning we have to manually connect pin 25 in the specific case of an lsdyna operator as it is not exposed as an available pin in the spec.
5. the way to gather the mesh for a given zone/part from a core.Model would need a helper as right now it basically returns the mesh_provider, without any way to get a meshes_provider nor give them any argument directly.
6. we require a mesh query API for the post.Simulation object.


rafacanton:
> This PR aims at improving the Core APIs regarding working with regions (zones, parts...).
> 
> Use Case	using core operators	using core helpers	using Post API
> Query the mesh for a given fluid zone	dpf.operators.mesh.mesh_provider(     data_sources=model,     region_scoping=single_zone_list, ).eval() --> DOES NOT ACCEPT A LIST OF ZONES	provider = model.metadata.mesh_provider provider.inputs.region_scoping.connect(single_zone_list) mesh = provider.eval() --> NOT A PROPER HELPER	Provide mesh query API -> add example
> Query a result on a given fluid zone	dpf.operators.result.static_pressure(     data_sources=model,     region_scoping=zone_list, ).eval()	model.results.static_pressure.on_region_scoping([4]).eval()	simulation.static_pressure(zone_ids=zone_list)
> Query the mesh for a given lsdyna part	dpf.operators.mesh.mesh_provider(     data_sources=model,     region_scoping=single_part_list, ).eval() --> ACCEPTS A LIST OF PARTS	provider = model.metadata.mesh_provider provider.inputs.region_scoping.connect(single_part_list) mesh = provider.eval() --> NOT A PROPER HELPER	Provide mesh query API -> add example
> Query a result for a given lsdyna part	op = dpf.operators.result.displacement(     data_sources=model, ) op.connect(25, part_list) fc = op.eval() --> non-fluid result operators do not have a region_scoping pin exposed	 	 
> Current observations:
> 
> 1. behavior of operator meshes_provider is not the same between fluid zones and lsdyna parts: for lsdyna the mesh is not labelled per part
> 2. behavior of operator mesh_provider is not the same between fluid zones and lsdyna parts: it accepts a list of parts for lsdyna and returns the merged mesh; it throws for multiple fluid zones
> 3. the region_scoping pin is not exposed for non-fluid operators, meaning we have to manually connect pin 25 in the specific case of an lsdyna operator as it is not exposed as an available pin in the spec.
> 4. the way to gather the mesh for a given zone/part from a core.Model would need a helper as right now it basically returns the mesh_provider, without any way to get a meshes_provider nor give them any argument directly.
> 5. we require a mesh query API for the post.Simulation object.

Hi, answering to the points above:

1. Correct, the plan is to update this server side, so the behavior is the same as in fluids (which is more aligned to the rest of Dpf).
2. Same as above, the mesh_provider output is a MeshedRegion, not a Container and as such it should only accept one id.
3. Correct, this needs to be addressed.
4. OK, and this is critical for performance. In my opinion, the Model should only query the MeshInfo to  make sure that it is light and it only extracts metadata, and only if the user wants to work with the mesh (to plot or query it), go through the mesh/meshes providers.
5. Agree

src/ansys/dpf/core/results.py
```diff
@@ -520,6 +527,52 @@ def on_mesh_scoping(self, mesh_scoping):
...
+        3
+        >>> fc[0].location
+        'Faces'
+
+        """
+        from ansys.dpf.core import locations
+        if isinstance(region_scoping, list):
+            region_scoping = Scoping(
+                ids=region_scoping,
+                location=locations.zone,
```
rafacanton, line 569: If the location "zone" is the default, this is only going to be useful for fluids.

src/ansys/dpf/core/results.py
```diff
@@ -520,6 +527,52 @@ def on_mesh_scoping(self, mesh_scoping):
         self._mesh_scoping = mesh_scoping
         return self
 
+    def on_region_scoping(self, region_scoping):
```
PProfizi, line 530:
```suggestion
    def on_region_scoping(self, region_scoping: Scoping|list[int], type: locations = locations.zone):
```

src/ansys/dpf/core/results.py
```diff
@@ -520,6 +527,52 @@ def on_mesh_scoping(self, mesh_scoping):
...
+        >>> model = dpf.Model(examples.download_fluent_mixing_elbow_steady_state())
+        >>> disp = model.results.displacement
+        >>> fc = disp.on_region_scoping([1,2,3]).eval()
+        >>> len(fc[0].scoping)
+        3
+
+        Use a scoping to specify a list of entity IDs with their locations.
+
+        >>> stress = model.results.stress
+        >>> scop = dpf.Scoping(ids=[3,4,5], location= dpf.locations.zone)
```
PProfizi, line 557:
```suggestion
        >>> scop = dpf.Scoping(ids=[3,4,5], location=dpf.locations.zone)
```

src/ansys/dpf/core/results.py
```diff
@@ -598,6 +605,59 @@ def on_mesh_scoping(self, mesh_scoping):
...
+        >>> stress = model.results.stress
+        >>> scop = dpf.Scoping(ids=[3,4,5], location=dpf.locations.zone)
+        >>> fc = stress.on_region_scoping(scop).eval()
+        >>> len(fc[0].scoping)
+        3
+        >>> fc[0].location
+        'Faces'
+
+        """
+        from ansys.dpf.core import locations
```
PProfizi, line 646:
```suggestion
        from ansys.dpf.core import locations

```

src/ansys/dpf/core/results.py
```diff
@@ -598,6 +605,60 @@ def on_mesh_scoping(self, mesh_scoping):
...
+
```
PProfizi, line 647:
```suggestion
```

PR #1353, opened 2024-01-16
fix deep copy

cbellot000:


anseramosfe (approved):

src/ansys/dpf/core/core.py
```diff
@@ -282,12 +282,15 @@ def _deep_copy(dpf_entity, server=None):
                                 core.Field, core.FieldsContainer, core.MeshedRegion...
     """
     from ansys.dpf.core.operators.serialization import serializer_to_string, string_deserializer
-    from ansys.dpf.core.common import types_enum_to_types
+    from ansys.dpf.core.common import types_enum_to_types, types
 
-    serializer = serializer_to_string(server=server)
+    serializer = serializer_to_string(server=dpf_entity._server)
```
rafacanton, line 287: @cbellot000 Why serializing in a different server in which later on we will deserialize? So the info that is passed through (potentially) gRPC is the binary serialized?

cbellot000: @rafacanton you have to serialize on the server where the data is, if not: you are serializing twice: once to move the data to the other server and once to get the string

rafacanton: Makes sense, thanks!

rafacanton (approved):

PR #1372, opened 2024-01-24
Create a ResultInfo

PProfizi:
Allow creation of a ResultInfo. Only available for InProcess for now.

### DIFF
--- a/src/ansys/dpf/core/result_info.py
+++ b/src/ansys/dpf/core/result_info.py
@@ -5,6 +5,7 @@ ResultInfo
 import traceback
 import warnings
 
+from typing import List, Union
 from enum import Enum, unique
 from types import SimpleNamespace
 from ansys.dpf.gate import (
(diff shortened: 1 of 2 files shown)


src/ansys/dpf/core/result_info.py
```diff
@@ -61,14 +65,25 @@ class ResultInfo:
 
     This class describes the metadata of the analysis and the available results.
 
+    .. note::
+      Creating a new ResultInfo from an analysis type and physics type is currently only available
```
cbellot000, line 69: I see that you add more more capabilities available witha single server type, I'm afraid that it's going to make the library hard to use :/ (I found the issue with the streams class 2 days ago)

cbellot000: Are you following up to expand to other server types after?

PProfizi: @cbellot000 definitely, yes. But this can be merged first to allow for a custom operator to create a ResultInfo.

cbellot000 (approved):

PR #1376, opened 2024-01-24
Allow GenericDataContainer as input/output of a CustomOperator and fix dpf-site.zip update

PProfizi:
Also fix some typehint and some imports.

### DIFF
--- a/src/ansys/dpf/core/_custom_operators_helpers.py
+++ b/src/ansys/dpf/core/_custom_operators_helpers.py
@@ -146,7 +146,7 @@ _type_to_input_method = [
     ),
     (
         generic_data_container.GenericDataContainer,
-        external_operator_api.external_operator_put_out_generic_data_container,
+        external_operator_api.external_operator_get_in_generic_data_container,
         "generic_data_container",
     )
     # TO DO : (dpf_operator.Operator, external_operator_api.external_operator_get_in_operator,
--- a/src/ansys/dpf/core/_custom_operators_helpers.py
+++ b/src/ansys/dpf/core/_custom_operators_helpers.py
@@ -144,6 +144,11 @@ _type_to_input_method = [
         external_operator_api.external_operator_get_in_data_tree,
         "data_tree",
     ),
+    (
+        generic_data_container.GenericDataContainer,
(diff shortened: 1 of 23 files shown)


src/ansys/dpf/core/_custom_operators_helpers.py
```diff
@@ -70,6 +70,10 @@ def __operator_main__(operator_functor, data):
         custom_type_field.CustomTypeField,
         external_operator_api.external_operator_put_out_custom_type_field,
     ),
+    (
```
cbellot000, line 73: @PProfizi is it tested somewhere else?

cbellot000 (approved):

src/ansys/dpf/core/_custom_operators_helpers.py
PProfizi: @cbellot000 CodeCov says yes, which is weird because it is not tested explicitly.

PProfizi: @cbellot000 BTW why do we have a `type_to_output` specific to the custom operators?

PProfizi: @cbellot000 I jsut realized that to be be able to test this (and for this to even work), the changes need to be propagated to `dpf-site.zip`.
So either I add the capability to update the `dpf-site.zip` based on the current venv as we discussed previously, and use it at the start of `test_python_plugins.py`,
or we have to merge it first, then wait for a server sync loop to put it in `dpf-site.zip` and only then add the tests.

PProfizi: See https://github.com/ansys/pydpf-core/pull/1379

cbellot000: I see then, you can maybe add the test ad mark it as skip this way we can remove the skip later?

cbellot000: > @cbellot000 BTW why do we have a `type_to_output` specific to the custom operators?

I don't understand, where would you put "external_operator_put_out_generic_data_container" instead?

PR #1381, opened 2024-01-26
Fix test_entry.py::test_license_agr

ansys-akarcher:
Following license changes in DPF Server, this test was failing

Basically the behavior should not be modified apart from the Operator Construction also throwing a DPF LA Exception.

However the test seemed to run an InProcessServer before even actually running the line which starts a server.

Which caused all plugins to be already loaded, and DPF to be already loaded, and I believe biasing the expected environnement of the test.

This should be merged in synchronization with the Licensing modifications server side

### DIFF
--- a/tests/entry/test_entry.py
+++ b/tests/entry/test_entry.py
@@ -8,6 +8,7 @@ from ansys.dpf.core import examples
 from ansys.dpf.core.core import errors
 from conftest import running_docker
 
+import time
 
 @pytest.mark.order(1)
 @pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


tests/entry/test_entry.py
```diff
@@ -8,6 +8,7 @@
 from ansys.dpf.core.core import errors
 from conftest import running_docker
 
+import time
```
PProfizi, line 11: @ansys-akarcher is this required?

tests/entry/test_entry.py
```diff
@@ -17,12 +18,13 @@
...
 )
 def test_license_agr(restore_accept_la_env):
+    dpf.server.shutdown_global_server()
     config = dpf.AvailableServerConfigs.InProcessServer
     init_val = os.environ["ANSYS_DPF_ACCEPT_LA"]
     del os.environ["ANSYS_DPF_ACCEPT_LA"]
     with pytest.raises(errors.DPFServerException):
         dpf.start_local_server(config=config, as_global=True)
-    with pytest.raises(KeyError):
+    with pytest.raises(errors.DPFServerException):
```
PProfizi, line 27: @ansys-akarcher How can this work for older server versions?

tests/entry/test_entry.py
```diff
@@ -17,12 +18,13 @@
     reason="Tests ANSYS_DPF_ACCEPT_LA",
 )
 def test_license_agr(restore_accept_la_env):
+    dpf.server.shutdown_global_server()
```
PProfizi, line 21: @ansys-akarcher can you confirm that this not being here before, was the reason why we got a `KeyError` as a second exception, and now that you shutdown the global server first, you do get a `DPFSErverException` as a second exception even for older servers? Is this expected?

PProfizi (approved):

tests/entry/test_entry.py
PProfizi, line 11:
```suggestion
```

ansys-akarcher: It is not

PR #1400, opened 2024-02-05
fix: ensure that Meilisearch secret is used in workflows

### LINKED ISSUE #1327: PyDPF documentation search doesn't work
### Description of the modifications

PyDPF documentation search doesn't work - it was resolved some time back but now it doesn't work.
The first search returns a blank page. Searching again on that blank page gives a cropped drop-down of results.

image

_No response_

jorgepiloto:
Continuation of #1302, fixes #1327 by adding the variable at step level. Everything is properly configured in the `conf.py` file and I can see the Meilisearch indices for PyDPF working as expected.

The latest stable documentation does not contain any reference to Meilisearch in its HTML, meaning that the `MEILISEARCH_PUBLIC_API_KEY` was not available when the docs built. I suspect that the `cmd` is starting a new shell with a new env, loosing track of the environment declared in the `docs.yml`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -44,7 +44,6 @@ on:
 env:
   PACKAGE_NAME: ansys-dpf-core
   MODULE: core
-  MEILISEARCH_PUBLIC_API_KEY: ${{ secrets.MEILISEARCH_PUBLIC_API_KEY }}
 
 jobs:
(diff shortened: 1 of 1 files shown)


PProfizi (approved):

MaxJPRey:
Thanks @jorgepiloto for fixing it.

PR #1445, opened 2024-02-28
Add a key argument to DataSources.set_domain_result_file_path()

PProfizi:


src/ansys/dpf/core/data_sources.py
```diff
@@ -135,18 +136,22 @@ def guess_result_key(filepath: str) -> str:
                 return result_key
         return ""
 
-    def set_domain_result_file_path(self, path, domain_id):
+    def set_domain_result_file_path(
+            self, path: Union[str, os.PathLike], domain_id: int, key: str = ""
```
FedericoNegri, line 140: minor: why not defaulting `key` to `None`, since you only use it if set

FedericoNegri (approved):
Tried out and it works well, thanks for the quick turnaround!

src/ansys/dpf/gate/data_sources_grpcapi.py
```diff
@@ -60,6 +60,18 @@ def data_sources_set_domain_result_file_path_utf8(dataSources, name, id):
         request.data_sources.CopyFrom(dataSources._internal_obj)
         _get_stub(dataSources._server).Update(request)
 
+    @staticmethod
+    def data_sources_set_domain_result_file_path_with_key_utf8(dataSources, name, key, id):
+        from ansys.grpc.dpf import data_sources_pb2
```
rafacanton, line 65: @PProfizi Shouldn't this be done in the place where this code is maintained?

rafacanton (changes requested):

src/ansys/dpf/gate/data_sources_grpcapi.py
PProfizi: @rafacanton yes, it will. This is to test the change. we can also merge it from here while waiting for it to come from the server-sync. It will be crushed anyway.

cbellot000 (approved):

src/ansys/dpf/gate/data_sources_grpcapi.py
cbellot000: I agree with @rafacanton, we should even gitignore this files because if you're not there to check that manually, then we can have problems: when updating dpf-standalone, these files will be erased and and we will have failing unit tests

PProfizi: @rafacanton @cbellot000 I propose that the `_grpcapi.py` files of `ansys/dpf/gate/` should be versioned in `pydpf-core` directly. Then all the server-sync does is provide a `ansys/dpf/gate/generated` folder to replace the ones in `pydpf-core`.
If we gitignore these files, then they will be ignored and not tracked.

src/ansys/dpf/gate/data_sources_grpcapi.py
```diff
@@ -60,6 +60,18 @@ def data_sources_set_domain_result_file_path_utf8(dataSources, name, id):
...
+        request = data_sources_pb2.UpdateRequest()
+        request.result_path = True
+        request.domain.domain_path = True
+        request.domain.domain_id = id
```
PProfizi, line 69:
```suggestion
    @staticmethod
    def data_sources_set_domain_result_file_path_with_key_utf8(dataSources, name, key, domain_id):
        from ansys.grpc.dpf import data_sources_pb2
        request = data_sources_pb2.UpdateRequest()
        request.result_path = True
        request.domain.domain_path = True
        request.domain.domain_id = domain_id
```

rafacanton (approved):

PR #1450, opened 2024-03-01
Fix Docker for 24.2.pre1

PProfizi:
The Docker jobs started failing on February 26 while they were not just moments before, most likely due to the weekly GH runners update that day.
In this case the deployment of their [new Ubuntu latest runner image](https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240225.1) updated the Docker packages.
This is still under investigation.


BClappe (approved):

PProfizi:
Closing as this went away on its own

PR #1462, opened 2024-03-08
refactor: Added types to operator and pin specifications

BClappe:
Added types to operator and pin specifications


src/ansys/dpf/core/operator_specification.py
```diff
@@ -272,7 +273,7 @@ def __str__(self):
         return "Description:\n" + str(self.description) + "\nProperties:\n" + str(self.properties)
 
     @property
-    def properties(self):
+    def properties(self) -> dict:
         """Returns some additional properties of the Operator, like the category, the exposure,
         the scripting and user names and the plugin
```
PipKat, line 278:
```suggestion
        """some additional properties of the Operator, like the category, the exposure,
        the scripting and user names, and the plugin
```
Properties should be described as nouns. We also use the Oxford comma, which is placing a comma before the "and" or the "or" in a list or more than two items.

src/ansys/dpf/core/operator_specification.py
```diff
@@ -68,7 +69,7 @@ def __init__(
         self.name_derived_class = name_derived_class
 
     @property
-    def type_names(self):
+    def type_names(self) -> list:
```
PProfizi, line 72:
```suggestion
    def type_names(self) -> list[str]:
``` 
list of strings?

src/ansys/dpf/core/operator_specification.py
```diff
@@ -103,7 +104,7 @@ def type_names(self, val):
         self._type_names = val
 
     @staticmethod
-    def _get_copy(other, changed_types):
+    def _get_copy(other, changed_types) -> "PinSpecification":
```
PProfizi, line 107: Why not the actual class?

src/ansys/dpf/core/operator_specification.py
```diff
@@ -231,7 +232,7 @@ class Specification(SpecificationBase):
     'result file path container, used if no streams are set'
     """
 
-    def __init__(self, operator_name=None, specification=None, server=None):
+    def __init__(self, operator_name: Union[str, None]=None, specification: Union["Specification", None]=None, server: Union[server_module.BaseServer, None]=None):
```
PProfizi, line 235: Again, you could use `Specification` directly by using `from typing import TYPE_CHECKING` and `if TYPE_CHECKING: import Specification`, you can see it being done in `GenericDataContainer.py`.
Also `from __future__ import annotations` as a first import

BClappe: I was afraid this could cause problems on runtime to our consumers, couldn't it?

src/ansys/dpf/core/operator_specification.py
BClappe: Because of imports problems. But they are fixed by importing annotations from __future__.

PProfizi (approved):

PR #1469, opened 2024-03-13
Add PropertyField.name

PProfizi:


src/ansys/dpf/core/property_field.py
```diff
@@ -101,6 +103,12 @@ def _field_create_internal_obj(
         else:
             return api.csproperty_field_new(nentities, nentities * dim.component_count)
 
+    def _load_field_definition(self):
+        """Attempt to load the field definition for this field."""
+        # try:
+        out = self._api.csproperty_field_get_shared_field_definition(self)
```
rafacanton, line 109: This will only exist for servers >= 2024.2.pre1

PProfizi: @rafacanton added decorators to raise an error if the server version is not sufficient

rafacanton (approved):

PR #1472, opened 2024-03-18
docs: update the style of the operator reference in docs

Revathyvenugopal162:
fix the issue with header and example in the operator.rst file in https://dpf.docs.pyansys.com/version/stable/operator_reference_load_apis.html image


PProfizi (approved):
Thanks @Revathyvenugopal162 , I can confirm it looks good now

PR #1473, opened 2024-03-18
Associate the mesh to the Nodes.coordinates field 

PProfizi:
This made the coordinates field unavailable for plotting and further handling by Post


src/ansys/dpf/core/nodes.py
```diff
@@ -225,7 +225,9 @@ def coordinates_field(self):
         DPFArray([0.015, 0.045, 0.03 ]...
 
         """
-        return self._get_coordinates_field()
+        cf = self._get_coordinates_field()
+        cf.meshed_region = self._mesh
```
rafacanton, line 229: Doesn't this create a sort of circular dependency? The mesh has a field inside (the coordinate field), that has the mesh as its support, that in turn has a coordinate field.... and so on

PProfizi: @rafacanton Notice that here I only do that when returning a "standalone" Field. Its associated mesh will indeed have a coordinates PropertyField, but that one will not have an associated mesh until we explicitly request it on its own

rafacanton: @PProfizi Yes, I understand that this association will not happen until this method is called. I cannot imagine of any potential downside that this can have down the stream (havbing created that circ dep from the time this association is made). @cbellot000 can you imagine of any?

PProfizi: Hi @rafacanton , Hi @cbellot000 ,
If we cannot think of a way this may break something, we might as well merge it.
It is mandatory for several purposes when handling a `PropertyField`.

rafacanton (approved):

PR #1494, opened 2024-04-10
Fix issue in model without faces

rafacanton:
A bug was reported server side pointing to a crash if faces are requested in a mesh that doesn't have them:
```
filename = examples.download_pontoon()
model = dpf.Model(filename)
mesh.faces.face_by_id(1) # server crashes since there are no faces
```
We now handle the situation here. A test is added to ensure that the raise is expected.

### DIFF
--- a/src/ansys/dpf/core/faces.py
+++ b/src/ansys/dpf/core/faces.py
@@ -296,8 +296,12 @@ class Faces:
         """
         if faceindex is None:
             faceindex = self._mesh.property_field(face_properties.faces_type).scoping.index(faceid)
+            if (faceindex < 0):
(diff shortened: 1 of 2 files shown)


PProfizi:
@rafacanton agreed to put this in place, yet shouldn't it be the server's job to return something empty instead of throwing? Food for thoughts

PProfizi:
> @rafacanton agreed to put this in place, yet shouldn't it be the server's job to return something empty instead of throwing? Food for thoughts

ah sorry, i just realized the issue here is that the client does not check the input it gives to the request.

PProfizi (approved):

PR #1514, opened 2024-04-19
Fix server shutdown

cbellot000:
BUG: calling shutdown on a server prevents from starting another server on linux.
-> when a server is started, to choose the port to start the server on, all server instances are requested their ports but shutdown prevents the server to return its port. FIX is to store the port of the server before shutdown

### DIFF
--- a/src/ansys/dpf/core/server_types.py
+++ b/src/ansys/dpf/core/server_types.py
@@ -382,6 +382,7 @@ class BaseServer(abc.ABC):
         self._session_instance = None
         self._base_service_instance = None
         self._context = None
+        self._info_instance = None
         self._docker_config = server_factory.RunningDockerConfig()
 
     def set_as_global(self, as_global=True):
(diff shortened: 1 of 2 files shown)


src/ansys/dpf/core/server_types.py
```diff
@@ -782,6 +784,7 @@ def _create_shutdown_funcs(self):
 
     def shutdown(self):
         if self.live:
+            info = self.info # initializing the info variable (giving access to ip and port): this can be required if start_local_server is called afterwards
```
cbellot000, line 787:
```suggestion
            _ = self.info # initializing the info variable (giving access to ip and port): this can be required if start_local_server is called afterwards
```

src/ansys/dpf/core/server_types.py
```diff
@@ -1221,14 +1210,15 @@ def local_server(self, val):
 
     def shutdown(self):
         if self._own_process and self.live:
+            info = self.info # initializing the info variable (giving access to ip and port): this can be required if start_local_server is called afterwards
```
cbellot000, line 1213:
```suggestion
            _ = self.info # initializing the info variable (giving access to ip and port): this can be required if start_local_server is called afterwards
```

PProfizi (approved):

PR #1526, opened 2024-04-22
MeshedRegion.plot() with PropertyField

PProfizi:
Add the possibility to plot a `PropertyField` on a `MeshedRegion`.
Will have to add `PropertyField.plot(mesh)` with mandatory mesh as a `PropertyField` currently most likely does not have a mesh asosciated.

The PyVista options for the scalar bar still need a little bit of tweaking. Some colors are even missing from the scalar bar:
image

### DIFF
--- a/src/ansys/dpf/core/plotter.py
+++ b/src/ansys/dpf/core/plotter.py
@@ -214,11 +214,18 @@ class _PyVistaPlotter:
         **kwargs,
     ):
         # Get the field name
-        name = field.name.split("_")[0]
-        unit = field.unit
-        kwargs.setdefault("stitle", f"{name} ({unit})")
-
-        kwargs = self._set_scalar_bar_title(kwargs)
+        if isinstance(field, dpf.core.Field):
+            name = field.name.split("_")[0]
+            categories = False
+            unit = field.unit
+            kwargs.setdefault("stitle", f"{name} ({unit})")
+            kwargs = self._set_scalar_bar_title(kwargs)
+        elif isinstance(field, dpf.core.PropertyField):
+            name = field.name
+            categories = True
+            kwargs.setdefault("stitle", f"{name}")
+            kwargs = self._set_scalar_bar_title(kwargs)
+            kwargs["scalar_bar_args"]["fmt"] = "%.0f"
 
         kwargs.setdefault("show_edges", True)
         kwargs.setdefault("nan_color", "grey")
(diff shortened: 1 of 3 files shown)


PProfizi:
Hi @AlejandroFernandezLuces, could you please take a look at this PR?
The goal is to plot a discrete and potentially discontinuous field of integers, each value colored.
Do you think the approach with `scalar_bar` and `categories` options is the right one?

src/ansys/dpf/core/plotter.py
```diff
@@ -214,11 +214,23 @@ def add_field(
...
+        elif isinstance(field, dpf.core.PropertyField):
+            try:
+                name = field.name
+            except dpf_errors.DpfVersionNotSupported:
+                name = ""
+            categories = True
+            kwargs.setdefault("stitle", f"{name}")
+            kwargs = self._set_scalar_bar_title(kwargs)
+            kwargs["scalar_bar_args"]["n_labels"] = len(set(field.data))
+            kwargs["scalar_bar_args"]["fmt"] = "%.0f"
```
AlejandroFernandezLuces, line 232: I would prefer to directly use the `add_scalar_bar` directly, I think something like this should work in this context
```suggestion
            self._plotter.add_scalar_bar(
                title='Your title',
                n_labels=len(set(field.data))
                fmt='%.0f',
)
```

AlejandroFernandezLuces (commented):
Yes, I think activating `categories` flag and using the `scalars_bar` is the best way to represent what you want. 

A side comment, why are you filling a `kwargs` map instead of using the arguments in the function calls? I personally find it clearer to use the parameters directly.

PProfizi:
Thanks @AlejandroFernandezLuces!
Also, any idea why a color is missing in the scalar bar of the plot I shared above?

AlejandroFernandezLuces:
> Thanks @AlejandroFernandezLuces! Also, any idea why a color is missing in the scalar bar of the plot I shared above?

Not sure what is happening there, you can try to set `n_colors=len(set(field.data))` in `add_scalar_bar` or in kwargs, so you make sure that you have a color per label on PyVista side. If not, maybe you have a bug on pyDPF that causes a misrepresentation between the fields and the meshes?

PProfizi:
NB: the default cmap used is `tab20` as it is discrete and ensures a clear difference between neighboring values, however it only has 20 values and will not "circle back" if more values are present. It will simply associate several consecutive values to one color.
Improvement should be made to create a dynamic color map based on semi-random colors, and to properly handle discontinuous discrete series of integers.

tests/conftest.py
```diff
@@ -19,10 +19,10 @@
 
 ACCEPTABLE_FAILURE_RATE = 0
 
-core.settings.disable_off_screen_rendering()
-os.environ["PYVISTA_OFF_SCREEN"] = "true"
-core.settings.bypass_pv_opengl_osmesa_crash()
-os.environ["MPLBACKEND"] = "Agg"
+# core.settings.disable_off_screen_rendering()
```
rafacanton, line 22: This needs to be reverted before merging, right?

PR #1527, opened 2024-04-22
Fix Operator.eval() in case output pin 0 is not named "fields_container"

PProfizi:
I have stumbled upon operators where the output pin 0, despite being of type `fields_container`, is not named `fields_container`. 
That is an implicit rule we generally follow, not a hard one we can expect everyone to follow.
The `eval()` method should thus be made more generic.
It still tests for a `fields_container` type output pin, but it does not put any requirement on the naming.

### DIFF
--- a/src/ansys/dpf/core/results.py
+++ b/src/ansys/dpf/core/results.py
@@ -263,7 +263,12 @@ class Result:
         >>> fc = disp.on_all_time_freqs.eval()
 
         """
-        fc = self.__call__().outputs.fields_container()
+        outputs = self.__call__().outputs
+        if hasattr(outputs, "fields_container"):
+            fc = outputs.fields_container()
+        else:
+            # Support operators where the first output is not named "field_container"
+            fc = outputs[0]()
         if self._specific_fc_type == "shape":
             fc = ElShapeFieldsContainer(fields_container=fc._get_ownership(), server=fc._server)
         elif self._specific_fc_type == "body":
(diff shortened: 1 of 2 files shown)


src/ansys/dpf/core/results.py
```diff
@@ -263,7 +263,12 @@ def eval(self):
         >>> fc = disp.on_all_time_freqs.eval()
 
         """
-        fc = self.__call__().outputs.fields_container()
+        outputs = self.__call__().outputs
+        if hasattr(outputs, "fields_container"):
+            fc = outputs.fields_container()
+        else:
+            # Support operators where the first output is not named "field_container"
+            fc = outputs[0]()
```
rafacanton, line 271: @PProfizi Would it be a very breaking change to make this the default? I mean, is the `if` part necessary, doesn't this also handle fields_container?

PProfizi: hey @rafacanton , osrry I just updated the description of the PR, you were too fast :)

PProfizi: actually @rafacanton  you're right, the logic in the else should also handle the first case... thanks for noticing that!

PProfizi, line 271:
```suggestion
        fc = self().outputs[0]()
```

PProfizi: ah wait... no it is not exactly the same thing @rafacanton , in the first case, if there is an output pin in any position called fields_container, we get this one. In the second case, we take the first output pin.
So I guess making this the default would not be retro-compatible.... sadly

rafacanton (approved):

cbellot000 (approved):

PR #1557, opened 2024-05-07
Launch server by version

rlagha:
Add a helper method which allows to generate new functions per installed DPF server to ease choosing which server to launch

### DIFF
--- a/src/ansys/dpf/core/server.py
+++ b/src/ansys/dpf/core/server.py
@@ -3,8 +3,10 @@ Server
 ======
 Contains the directives necessary to start the DPF server.
 """
+import functools
 import os
 import socket
+import sys
 import weakref
 import copy
 import platform
@@ -388,3 +390,46 @@ def get_or_create_server(server: BaseServer) -> Union[BaseServer, None]:
     if server:
         return server
     return _global_server()
+
+
+def available_servers():
+    """Searches all available installed DPF servers in the current machine
+    Binds new functions to the server module which helps to choose the appropritae/suitable version
+
+    after this call, you can do the following:
+
+    Examples
+    --------
+
+    >>> from ansys.dpf import core as dpf
+    >>> #out = dpf.server.available_servers()
+
+    launch a server.
+    >>> #server = dpf.server.start_2024_2_server()
+
+
+    Parameters
+    ----------
+    server: BaseServer, None
+
+    Returns
+    -------
+    server: returns the map of available DPF servers.
+    """
+    from ansys.dpf.gate import load_api
+    unified = load_api._paths_to_dpf_in_unified_installs()
+    standalone = load_api._paths_to_dpf_server_library_installs()
+
+    strver = {}
+
+    out = {}
+    strver.update(unified)
+    strver.update(standalone)
+    for version, path in strver.items():
+        bound_method = start_local_server
+        method2 = functools.partial(bound_method, ansys_path=path)
+        vout = str(version).replace(".", "_")
+        setattr(sys.modules[__name__], "start_" + vout + "_server", method2)
+        out[str(version)] = method2
+
+    return out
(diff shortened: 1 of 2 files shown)


src/ansys/dpf/core/server.py
```diff
@@ -388,3 +390,46 @@ def get_or_create_server(server: BaseServer) -> Union[BaseServer, None]:
     if server:
         return server
     return _global_server()
+
+
+def available_servers():
+    """Searches all available installed DPF servers in the current machine
+    Binds new functions to the server module which helps to choose the appropritae/suitable version
+
+    after this call, you can do the following:
```
PipKat, line 399:
```suggestion
    """Searches all available installed DPF servers on the current machine.
    
    This method binds new functions to the server module, which helps to choose the appropriate version.

    After calling this method, you can do the following:
```
There is no list of things you can do? Is this just not completed yet?

src/ansys/dpf/core/server.py
```diff
@@ -388,3 +390,46 @@ def get_or_create_server(server: BaseServer) -> Union[BaseServer, None]:
...
+    >>> #server = dpf.server.start_2024_2_server()
+
+
+    Parameters
+    ----------
+    server: BaseServer, None
+
+    Returns
+    -------
+    server: returns the map of available DPF servers.
```
PipKat, line 417:
```suggestion
    server: map of available DPF servers.
```
Because this is the "Returns" section, we don't need the "Returns the" part in the description for ``server``.

tests/test_server.py
```diff
@@ -316,6 +317,23 @@ def version(self):
 
     print(MockServer(remote_server).version)
     with pytest.raises(
-        ValueError, match="Error connecting to DPF LegacyGrpcServer with version 1.0"
+            ValueError, match="Error connecting to DPF LegacyGrpcServer with version 1.0"
```
PipKat, line 320: A messages should always have a verb and concluding period. They should also be in plain English. Something like this would be better: "Could not connect to DPF legacy gRPC server with version 1.0."

tests/test_server.py
```diff
@@ -316,6 +317,23 @@ def version(self):
...
     ):
         dpf.core.server_types.check_ansys_grpc_dpf_version(MockServer(remote_server), timeout=2.0)
+
+@pytest.mark.skipif(
+    os.name == "posix",
+    reason="can only work with local DPF server install",
+)
+def test_available_servers():
+    out = {}
+    out = server.available_servers()
```
cbellot000, line 330: @rlagha shouldn't you test if all the values of  `out` can be called?

cbellot000 (approved):

tests/test_server.py
```diff
@@ -316,6 +318,25 @@ def version(self):
...
+    for version in out:
+        vout = str(version).replace(".", "_")
+        if packaging.version.parse(version) > packaging.version.parse("2022.1"):
+            meth = "start_" + vout + "_server"
+            assert (hasattr(server, meth))
+            start = getattr(server, meth)
+            srv = start()
+            assert srv.local_server
+            srv = out[version]()
+            assert srv.local_server
```
PProfizi, line 342: Hey @rlagha,
thanks for the helper, this is great. I guess we should also test that the server version ran is the correct one.
```suggestion
            assert srv.local_server
            assert srv.version == version
            srv = out[version]()
            assert srv.local_server
            assert srv.version == version
```

tests/test_server.py
```diff
@@ -316,6 +318,25 @@ def version(self):
 
     print(MockServer(remote_server).version)
     with pytest.raises(
-        ValueError, match="Error connecting to DPF LegacyGrpcServer with version 1.0"
+            ValueError, match="Could not connect to DPF legacy gRPC server with version 1.0."
```
PProfizi, line 321: @rlagha why did you change the expected error?

PProfizi, line 321: This needs to be reverted as this is only what we expect from the server. Please change it server side if it needs changing.

tests/test_server.py
rlagha: @PipKat , thanks @PipKat , we will fix this in the server side.

PProfizi (approved):

PR #1559, opened 2024-05-10
Changes to support for 25.1 revision

guptadivya1501:


PProfizi:
Hi @guptadivya1501, I think the changes required to support `25.1` were already made.

guptadivya1501:
@PProfizi , If you check, it is merged yet.
image

src/ansys/dpf/core/_version.py
```diff
@@ -23,7 +25,7 @@ class ServerToAnsysVersion:
         "7.1": "2024R1",
         "8.0": "2024R2",
         "8.1": "2024R2",
-        "8.2": "2024R2",
```
BClappe, line 26: Why do we remove this line? @guptadivya1501

src/ansys/dpf/core/_version.py
```diff
@@ -23,7 +25,7 @@ class ServerToAnsysVersion:
...
+        "9.0": "2025R1",
```
PProfizi, line 28:
```suggestion
        "8.2": "2024R2",
        "9.0": "2025R1",
```

src/ansys/dpf/core/_version.py
guptadivya1501: @BClappe , Oops I dont know how it is removed. I will add again

PProfizi (approved):

PR #1600, opened 2024-05-31
Add plugins.py module with load_plugin_on_server helper

PProfizi:
This PR proposes a new module called `plugins` aimed at management of DPF Python plugins.

Here the helper `load_plugin_on_server` allows to automate the steps required to send a Python plugin server-side and load it correctly.

It is based on the idea that DPF Python plugins are shipped as Python libraries to install in the active Python virtual environment.

### DIFF
--- a/src/ansys/dpf/core/plugins.py
+++ b/src/ansys/dpf/core/plugins.py
@@ -5,7 +5,10 @@ Python DPF plugins utilities
 Contains the utilities specific to installing and using Python DPF plugins.
 """
 import os.path
-import importlib.metadata as importlib_metadata
+try:
+    import importlib.metadata as importlib_metadata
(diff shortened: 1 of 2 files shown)


ansys-akarcher (approved):

src/ansys/dpf/core/plugins.py
```diff
@@ -0,0 +1,77 @@
+"""
+Python DPF plugins utilities
+============================
+
+Contains the utilities specific to installing and using Python DPF plugins.
+"""
+import os.path
+import importlib.metadata as importlib_metadata
```
PProfizi, line 8:
```suggestion
try:
    import importlib.metadata as importlib_metadata
except ImportError:  # Python < 3.10 (backport)
    import importlib_metadata as importlib_metadata
```

PR #1608, opened 2024-06-04
Get package version from library metadata

PProfizi:
This PR implements retrieval of the package's version from its metadata, as defined in the `pyproject.toml`, as is shown in the PyAnsys best practices.

### DIFF
--- a/src/ansys/dpf/core/__init__.py
+++ b/src/ansys/dpf/core/__init__.py
@@ -1,8 +1,9 @@
 import os
 import pkg_resources
 
-from ansys.dpf.core._version import __version__
+import importlib.metadata as importlib_metadata
 
+__version__ = importlib_metadata.version(__name__.replace(".", "-"))
 
 # Setup data directory
 USER_DATA_PATH = None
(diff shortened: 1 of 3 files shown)


BClappe (approved):

src/ansys/dpf/core/__init__.py
```diff
@@ -1,8 +1,12 @@
 import os
 import pkg_resources
 
-from ansys.dpf.core._version import __version__
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:  # Python < 3.10 (backport)
+    import importlib_metadata as importlib_metadata
 
+__version__ = importlib_metadata.version("ansys-dpf-core")
```
BClappe, line 9: Where does it get the version from exactly? @PProfizi

BClappe (approved):

src/ansys/dpf/core/__init__.py
PProfizi: @BClappe from the `pyproject.toml`, meaning we have only one source of truth, and it is in the package metadata file.

PProfizi: This is part of the work to adhere to PyAnsys best practices.

PR #1613, opened 2024-06-13
Add a no_context ServerContext

PProfizi:
This `ServerContext` is interpreted so that the initialization is made without applying a context.


cbellot000 (approved):
Thank you so much for that Paul

cbellot000 (approved):

PR #1620, opened 2024-06-20
Add official support for CFX .res files

PProfizi:
The status of support for CFX `.res` files has been moved from unsupported to officially supported last week.
This PR intends to expose the capability of reading `.res` files from CFX. 

PyDPF-Post tests are failing due to the change to the `download_cfx_mixing_elbow` output.
Still, I'd rather switch to having it return a single path to the `.res` file and update the test in PyDPF-Post.


PProfizi:
PyDPF-Post tests are failing due to the change to the `download_cfx_mixing_elbow` output.
Still, I'd rather switch to having it return a single path to the `.res` file and update the test in PyDPF-Post.

rafacanton (approved):

PR #1626, opened 2024-06-21
Bump pyvista from 0.36.1 to 0.43.8 in /requirements

PProfizi:
Also add a step in docs.yml to install `open-mesa` for `vtk`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -117,7 +117,7 @@ jobs:
         run: python -c "from ansys.dpf import core"
 
       - name: "Setup headless display"
-        uses: pyvista/setup-headless-display-action@v1
+        uses: pyvista/setup-headless-display-action@v2
 
       - name: "Setup Graphviz"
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -133,6 +133,11 @@ jobs:
       - name: "Kill all servers"
         uses: ansys/pydpf-actions/kill-dpf-servers@v2.3
 
+      - name: "Ensure VTK compatibility"
+        run: |
+          python -m pip uninstall --yes vtk
+          python -m pip install --extra-index-url https://wheels.vtk.org vtk-osmesa==9.2.20230527.dev0
```
jorgepiloto, line 139: I will push one last commit such that this version is an ENV variable. This way, it will be easier for us to update it.

jorgepiloto: Done!

jorgepiloto (approved):

PR #1650, opened 2024-07-09
comment the docstring that is failing du to new element types (INTER and MPC)

MichaelNale:
comment docstring

### DIFF
--- a/src/ansys/dpf/core/results.py
+++ b/src/ansys/dpf/core/results.py
@@ -430,8 +430,8 @@ class Result:
         # <BLANKLINE>
         # ...1, 5, 6, 10, 2, 7, 8, 13, 4, 12, 15]...
         # >>> disp_mat_10 = fc_disp.get_field_by_mat_id(10)
-        #
-        # """
+
+        """
         self._specific_fc_type = "body"
         return self._add_split_on_property_type("mat")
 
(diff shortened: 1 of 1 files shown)


src/ansys/dpf/core/results.py
```diff
@@ -419,19 +419,19 @@ def split_by_body(self):
...
+        # >>> disp = model.results.displacement
+        # >>> fc_disp = disp.split_by_body.eval()
+        # >>> len(fc_disp)
+        # 11
+        # >>> fc_disp.get_mat_scoping().ids
+        # <BLANKLINE>
+        # ...1, 5, 6, 10, 2, 7, 8, 13, 4, 12, 15]...
+        # >>> disp_mat_10 = fc_disp.get_field_by_mat_id(10)
+        #
+        # """
```
anseramosfe, line 434: I guess this needs to be uncommented to mark the comment termination no?

MichaelNale: Yes!

anseramosfe (approved):

anseramosfe (commented):
I would link here the the issue with the task to uncomment the docstring so we do not forget :D

rafacanton (approved):

PR #1666, opened 2024-07-26
Refactor vtk_helper.py to allow appending data

PProfizi:
Add helpers to append data to an `UnstructuredGrid`:
- append_field_to_grid
- append_fields_container_to_grid


abhishekchitwar (commented):
@PProfizi, Great! LGTM! It would be great to have an option to update the name of field data in case of the dpf_fieldscontainer_to_vtk.
Thanks!

abhishekchitwar (approved):
Thanks!

PR #1672, opened 2024-07-31
Remove unnecessary busy threads reading from processes stdout/stderr

ansys-akarcher:
this PR does not fix anything in particular.

From looking at the code it seems like, when a gRPC server is created, two threads will be started until the server dies, especially if the server startup is successful.

I think the behavior can be kept without the existence of these two threads, as the general approach is: block until the server startup is successful, or an error is detected, or server has not started and time has ran out.

Current modification do not work with server within docker context.

### DIFF
--- a/src/ansys/dpf/core/server_types.py
+++ b/src/ansys/dpf/core/server_types.py
@@ -127,36 +127,32 @@ def _run_launch_server_process(
 def _wait_and_check_server_connection(
     process, port, timeout, lines, current_errors, stderr=None, stdout=None
 ):
-    if not stderr:
-
-        def read_stderr():
-            with io.TextIOWrapper(process.stderr, encoding="utf-8") as log_err:
-                for line in log_err:
-                    LOG.debug(line)
-                    current_errors.append(line)
-
-        stderr = read_stderr
-        # check to see if the service started
-    if not stdout:
-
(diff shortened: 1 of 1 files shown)


src/ansys/dpf/core/server_types.py
```diff
@@ -127,36 +127,32 @@ def _run_launch_server_process(
...
-        # print(lines)
-        started = any("server started" in line for line in lines)
-
+        ## STDOUT parsing
+        # Block for stdout output
+        stdout_line = process.stdout.readline().decode("utf-8")
+
+        # Detect when server is started
+        if stdout_line != None and "server started" in stdout_line:
+            started = True
```
PProfizi, line 140: @ansys-akarcher shouldn't we also break?

ansys-akarcher: Yeah I was wondering about that also

I don't know if it would be possible to have "server started" and some stderr output

in this case the behavior would be different (breaking would wrongly set started as True, skipping any errors)

PProfizi:
Hi @ansys-akarcher is this still ongoing work?

ansys-akarcher:
@PProfizi the changes were insufficient, as there was also the docker context to take into account.

But yeah I'd like to complete these changes at some point

PR #1685, opened 2024-08-05
Remove "if: always()" from tests.yml and bump all pydpf-actions to v2.3

PProfizi:
Also switch to artifact-upload@v4 and fix concurrent artifact upload

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -215,7 +215,6 @@ jobs:
         shell: pwsh
         run: |
           .github\workflows\scripts\separate_long_core_tests.ps1
-        if: always()
 
       - name: "Set pytest arguments"
         shell: bash
@@ -228,11 +227,9 @@ jobs:
         working-directory: tests
         run: |
           pytest $DEBUG $COVERAGE $RERUNS --junitxml=junit/test-results.xml  .
-        if: always()
 
       - name: "Kill all servers"
         uses: ansys/pydpf-actions/kill-dpf-servers@v2.3
-        if: always()
 
       - name: "Test API test_launcher"
         uses: nick-fields/retry@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -361,15 +346,13 @@ jobs:
...
-        uses: ansys/pydpf-actions/kill-dpf-servers@v2.2
-        if: always()
+        uses: ansys/pydpf-actions/kill-dpf-servers@v2.3
 
       - name: "Upload Test Results"
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: ${{ env.PACKAGE_NAME }}_${{ matrix.python-version }}_${{ matrix.os }}_pytest_${{ inputs.ANSYS_VERSION }}
+          name: ${{ env.PACKAGE_NAME }}_${{ matrix.python-version }}_${{ matrix.os }}_pytest_${{ inputs.ANSYS_VERSION }}_${{ inputs.test_any == 'true' && 'any' || '' }}
```
BClappe, line 354: If this condition concatenates string (which I don't know), in some cases there will be a dangling underscore. Is that intended?
Did you also test that it concatenates strings as you most probably want?
```suggestion
          name: ${{ env.PACKAGE_NAME }}_${{ matrix.python-version }}_${{ matrix.os }}_pytest_${{ inputs.ANSYS_VERSION }}${{ inputs.test_any == 'true' && '_any' || '' }}
```

BClappe (approved):

.github/workflows/tests.yml
PProfizi: @BClappe yes, It does indeed concatenates at the right time. I did miss the trailing underscore though, thanks.

Issue #1644, opened 2024-06-28
Add deformation to the mode shape animation


Rfaelv:
### Description of the feature

The actual mode shape animation looks like [this example](https://dpf.docs.pyansys.com/version/stable/_images/sphx_glr_01-plot_and_animate_modes_007.png) presented in the documentation. Even though the method `animate_mode` allows the `deform_scale_factor` parameter, only the colors changes in the animation. I expected that the node coordinates changes too (acording to the mode shape).

If some of you could please implement that feature I would apreciate it so much.

Rfaelv:
I've driven into the code and found a way (not necessarily the better) to show the mode shape deformation. Below I describe the solution:

Here: https://github.com/ansys/pydpf-core/blob/05399b10cd293c63020ea224d13d3b3dba1f11d4/src/ansys/dpf/core/animation.py#L98 I added an addtional workflow output for `deform_by` as follows:
`wf.set_output_name("deform_by", scaling_op.outputs.field)`


Finally, in method https://github.com/ansys/pydpf-core/blob/05399b10cd293c63020ea224d13d3b3dba1f11d4/src/ansys/dpf/core/animator.py#L98 I passed the parameter `scale_factor` with the following value: `scale_factor[frame]`

PR #1696, opened 2024-08-13
Fixes the mesh deformation for mode shape animation

rajesh1359:
This commit make deform_by default for mode animation.

Fixes issue https://github.com/ansys/pydpf-core/issues/1032


landon-kanner:
In my testing, this branch fixes https://github.com/ansys/pydpf-core/issues/1644, but does not fix https://github.com/ansys/pydpf-core/issues/1032.

The mode shapes now deform but I cannot control the deformation scaling.

PProfizi (approved):

PR #1725, opened 2024-09-03
Fix label space string representation

PProfizi:


src/ansys/dpf/core/label_space.py
```diff
@@ -46,10 +46,22 @@ def _data_processing_core_api(self):
         core_api.init_data_processing_environment(self)
         return core_api
 
-    def fill(self, label_space):
+    def fill(self, label_space: dict):
```
BClappe, line 49: By default this is `dict[Any,Any]`. Do you have a more specific type?

src/ansys/dpf/core/label_space.py
```diff
@@ -46,10 +46,22 @@ def _data_processing_core_api(self):
...
         for key, index in label_space.items():
             self._api.label_space_add_data(self, key, index)
 
+    def __str__(self):
+        return str(dict(self))
```
BClappe, line 54: Is this really good practice?
Maybe you should consider `__repr__` [instead](https://medium.com/@akshatgadodia/understanding-pythons-str-and-repr-methods-a-guide-for-python-developers-83490d83e4e9#:~:text=__str__%20focuses%20on,relevant%20information%20for%20object%20reconstruction.)?

tests/test_label_space.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_label_space_(server_type):
+    reference = {"test": 1, "various": 2}
+    ls = dpf.LabelSpace(label_space=reference)
+    assert dict(ls) == reference
+    assert str(ls)
+    reference = {"test": 1, "various": 2}
+    ls.fill(label_space=reference)
+    print(ls)
```
BClappe, line 11: Maybe you should not leave a `print` in the test but test the changes that you did right above
```suggestion
```

src/ansys/dpf/core/label_space.py
PProfizi: @BClappe well, this is __str__, so I'd expect a user-friendly and readable string, such as provided by str().

tests/test_label_space.py
PProfizi: Yes sorry that's a rogue forgotten print

BClappe (approved):

PR #1730, opened 2024-09-04
Take ownership of ansys/dpf/gate/*grpcapi.py files

PProfizi:
A first PR to stop updating `ansys/dpf/gate/*_grpcapi.py` files when getting code updates from new servers.
This PR means that PyDPF-Core would now have ownership of these files.

Example run of code update:
https://github.com/ansys/pydpf-core/actions/runs/10698323848/job/29657450211

### DIFF
--- a/.ci/update_dpf_dependencies.py
+++ b/.ci/update_dpf_dependencies.py
@@ -66,10 +66,9 @@ if gate_path is not None:
     print(f"from {dist_path}")
     destination = os.path.join(core_path, "src", "ansys", "dpf", "gate", "__init__.py")
     print(f"into {destination}")
-    shutil.copytree(
+    shutil.copy(
         src=dist_path,
         dst=destination,
-        dirs_exist_ok=True,
     )
     print("Done updating ansys.dpf.gate generated code")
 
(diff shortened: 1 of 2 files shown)


rafacanton (approved):

.ci/update_dpf_dependencies.py
```diff
@@ -45,24 +45,33 @@
...
-    print("Done updating ansys-grpc-dpf")
+    print("Done updating ansys.grpc.dpf")
 else:
     print(f"{grpc_path_key} environment variable is not defined. " "Cannot update ansys-grpc-dpf.")
 
 if gate_path is not None:
     # Update ansys-dpf-gate
-    print("Updating ansys.dpf.gate")
-    dist_path = os.path.join(gate_path, "ansys-dpf-gate", "ansys")
+    print("Updating ansys.dpf.gate generated code")
```
PProfizi, line 54: @rafacanton @BClappe the TFS CI uses this script here to update ansys-dpf-core based on ansys-dpf-gate, so the issue was that this change is not yet merged. I'll test it and keep you updated.

PProfizi: Can confirm this now works on TFS

Issue #1722, opened 2024-09-02
Modernize the PyDPF repositories to PyAnsys standards


PProfizi:
### Discussed in https://github.com/ansys/pydpf-core/discussions/1697

<div type='discussions-op-text'>

<sup>Originally posted by **jorgepiloto** August 14, 2024</sup>
The current pipelines of the project are unmaintainable. This creates a bottleneck for merging pull-requests and solving issues. The branch strategy is not compliant with PyAnsys guidelines and the docs layout can be improved.

Breaking changes are required (and worth) for modernizing this project:

1. Generate Dockerfiles, build images, and upload these to the private GitHub Container Registry. These are only accessible by Ansys employees and authorized CI/CD runs.

2. Use Docker images in CI/CD. Limit the number workflows to: `ci_cd_pr.yml`, `ci_cd_main.yml`, `ci_cd_night.yml`, and `ci_cd_release.yml`. These provide the same capabilities as the current ones while being optimized. Use `labels` for skipping jobs.

3. Take advantage of matrix strategy to simplify backwards compatible testing.

4. Rename `master` to `main`. Use `release/X.Y` branches rather than `release/X.Y.Z`

5. Use the latest documentation theme. Use a question-card strategy for navigation. Enable sphinx-autoapi for improving the API reference. Use Sphinx hooks for custom doc-build behavior.

6. Clean project's root directory. Unnecessary files: `Makefile`, `testing.md`, `run_client.bat`, `ignore_words.txt`, `.flake8`, `pytest.ini`.

Implementing these tasks is not easy. We all need to agree and commit to these points. It requires weeks of work and coordination between teams.

The final result is a project with optimized pipelines, maintainable workflows, and amazing docs that allows developers to focus on actual project development.</div>

PR #1736, opened 2024-09-09
feat: add license headers

jorgepiloto:
This pull-request implements the license headers as part of the #1722.


.pre-commit-config.yaml
```diff
@@ -31,6 +31,13 @@ repos:
   - id: debug-statements
   - id: requirements-txt-fixer
 
+- repo: https://github.com/ansys/pre-commit-hooks
+  rev: v0.4.3
+  hooks:
+    - id: add-license-headers
+      args:
+      - --start_year=2020
```
jorgepiloto, line 39: Adding this as the start year after taking a look to the history of the `LICENSE` file, see https://github.com/ansys/pydpf-core/commits/master/LICENSE

PProfizi (approved):

jorgepiloto:
Apparently, there is no way to ignore the global `exclude` in pre-commit, see https://stackoverflow.com/questions/69106270/can-i-exclude-files-from-every-pre-commit-hook-except-for-one-without-repeating

We can change the exclude to a per-hook basis if we want the license headers to apply to all files within the `src/` dir.

PR #1737, opened 2024-09-09
Comment some tests due to change in average_on_label_fc

MichaelNale:
Due to fixes on average_on_label_fc, have to update tests in pydpf core


PProfizi:
@MichaelNale can't we update the refs instead of commenting the tests out ?

rafacanton:
> @MichaelNale can't we update the refs instead of commenting the tests out ?

We normally do this in a 2 step process: first comment the tests so server CI can pass, then move the server to GH and then uncomment the tests with different references for old and new server versions

PProfizi (approved):

PR #1746, opened 2024-09-16
Remove a test due to changes on eCONTA174 (consider as linear/quadratic instead of semi parabolic)

MichaelNale:
Change server side the eleType related to eCONTA174. We used to consider it as semi parabolic if he was surrounded by linear elements and that lead to inconsistency (for example to compute element::volume). Now, regarding this, we linearize this element or keep it quadratic.


rafacanton:
@MichaelNale I guess that when the changes are ported to core, you will create a new test that tests these functionalities, right?

MichaelNale:
@rafacanton At the end I can just update to reflect the fact that these elements are not semi parabolic anymore. Here it will be just `assert dpf.core.element_types.descriptor(el.type).n_nodes = len(el.connectivity)`.

rafacanton:
> @rafacanton At the end I can just update to reflect the fact that these elements are not semi parabolic anymore. Here it will be just `assert dpf.core.element_types.descriptor(el.type).n_nodes = len(el.connectivity)`.

I was referring to the opposite. Instead of changing this test to ensure that this element is no longer semiparabolic, you can also find a proper semiparabolic element in another model and test the same assertion (and that way you are still covering the scenario of this test!)

rafacanton (approved):

PR #1747, opened 2024-09-16
Fix coordinate system example retro-compatibility

PProfizi:
The modification made [here](https://github.com/ansys/pydpf-core/pull/1738/files) to follow the changes server-side were not retro-compatible with older DPF versions.


examples/10-mesh_operations/13-nodes_in_local_coordinate_system.py
```diff
@@ -55,7 +55,12 @@
 ###############################################################################
 # Get the rotation matrix of the LCS ID 12.
 # The first 9 values in the ``cs`` output is the rotation matrix.
-cs = dpf.operators.result.coordinate_system()
+try:
+  # Starting with DPF 2025.1.pre1
+  cs = dpf.operators.result.coordinate_system()
+except KeyError:
+  # For previous DPF versions
+  cs = model.operator(r"mapdl::rst::CS")
```
oparreno, line 63: I don't think you need to load the data sources input pin if you use the model approach.

oparreno (approved):

PR #1749, opened 2024-09-16
Add tutorials and update User Guide layout

luisaFelixSalles:
This PR changes the general layout of the user guide documentation section 

- Update the Ansys sphinx theme
- Add subtitles in the sidebar 
- Transfer the concepts section into the user guide section 
- Introduce the concept of Tutorials


doc/source/conf.py
```diff
@@ -193,7 +194,7 @@ def reset_servers(gallery_conf, fname, when):
 # -- Options for HTML output -------------------------------------------------
 html_short_title = html_title = "PyDPF-Core"
 html_theme = "ansys_sphinx_theme"
-html_logo = pyansys_logo_black
+html_logo = pyansys_logo_light_mode
```
PProfizi, line 197: @luisaFelixSalles why do we need this change? Is it going to work in dark mode?

doc/source/user_guide/main_entities.rst
```diff
@@ -1,14 +1,13 @@
...
 
-.. toctree::
-   :hidden:
+The following tutorials shows in details how to use basic features of PyDPF-Core
+for each usage main step.
 
-   model
-   operators
-   fields_container
+Define simulation data
```
PProfizi, line 9:
```suggestion
Define the origins of the data
```

doc/source/conf.py
luisaFelixSalles: The current nav bar looks like this:

image

 The logo has a black square background, but we dont see it because the navigation bar also has a black background.

So, I changed the logo for the one with a light back ground, to match the new navigation bar color. It actually doesn't look good in dark mode, I'll look into other projects to see how they do it

luisaFelixSalles: @Revathyvenugopal162 I would like to ask for your help also in this case:

I tried passing the logo as an html configuration option with:
image

But I get the following error when building the documentation: 

image

Do you know if I'm missing other configuration changes in other to solve this problem?

Revathyvenugopal162: This wont work with out new theme. Please add this "logo": "pyansys" in theme options. example PR , https://github.com/ansys/pyansys-geometry/commit/40366c1dfbdc13664d9c9448a970819a0990978e#diff-d8d3ed25802824d15bf411f8e97416d8fc6f9247e821b0617f2b869dc584b99c

luisaFelixSalles: I tried this but it gives me the same error. 
Is there a way to implement the logo (that respects the light and dark modes) with the new theme?

Revathyvenugopal162, line 197:
```suggestion
```
yes you need to remove this line, and then add "logo":"pyansys", in html theme options. That should work

luisaFelixSalles: I just test this variation and it still gives me the same error

Revathyvenugopal162: let me try it locally.

Revathyvenugopal162: yes, you are correct, and i think it is because of some mismatch in configuration. i can fix with with another workaround and push that changes if its ok for you

Revathyvenugopal162: @luisaFelixSalles @PProfizi The logo issue was fixed in [33fade6](https://github.com/ansys/pydpf-core/pull/1749/commits/33fade68ec366d400711f080eceec6f563fd789d) and this is the first time we are getting this issue. i will keep track on this and try to fix it in ansys sphinx theme if required

luisaFelixSalles: Thank you for your help ! 🙂

PR #1752, opened 2024-09-17
changes in the sidebar of the examples section

luisaFelixSalles:
This PR changes the sidebar in the examples web page:

- Add subtitles
image

### DIFF
--- a/doc/source/_templates/custom_examples.html
+++ b/doc/source/_templates/custom_examples.html
@@ -1,0 +1,36 @@
+<div class="sidebar-primary-items__start sidebar-primary__section">
+        <div class="sidebar-primary-item">
+<nav class="bd-docs-nav bd-links" aria-label="Section Navigation">
+  <p class="bd-links__title" role="heading" aria-level="1">Section Navigation</p>
+  <div class="bd-toc-item navbar-nav">
+    <p style="font-weight:bold;" aria-label="Fundamentals" role="heading" aria-level="2">Fundamentals</p>
+    <ul class="nav bd-sidenav">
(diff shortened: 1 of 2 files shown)


luisaFelixSalles:
@Revathyvenugopal162 I'm having a problem with changing the sidebar in this section as we use the sphinx gallery to generate the examples webpage. @PProfizi suggested asking you if you could help me here.

I did an html template in order to change this side bar. You can see the code in "doc/source/_templates/custom-examples.html".

In my conf.py file I use this configuration:
image

I tried multiples variations for the path pattern but I always get a problem when trying to browse through different subsections without returning to the main examples page. 

For example: 

if I go to "...examples/02-modal-analyses/index.html" 
image

and then I go to "...examples/02-modal-analyses/03-harmonic_analyses/index.html" I get this error:
image

Thanks for your help

Revathyvenugopal162:
@luisaFelixSalles Thanks for reaching out. I think it is an issue with the subdirectories. Can you please try "examples/*/**" in the conf.py file. If that is not solving the issue, may be we can talk in private and try to debug.

luisaFelixSalles:
> @luisaFelixSalles Thanks for reaching out. I think it is an issue with the subdirectories. Can you please try "examples/*/**" in the conf.py file. If that is not solving the issue, may be we can talk in private and try to debug.

I just tried this configuration but it doesn't work neither. Also my first examples page (...examples/index.html) doesn't use the template. 
image

Revathyvenugopal162:
> @Revathyvenugopal162
> 
> I just tried this configuration but it doesn't work neither. Also my first examples page (...examples/index.html) doesn't use the template. image

ok, got it.i will try this locally and let you know.

PR #1783, opened 2024-10-01
Fix minimal required protobuf

PProfizi:
Due to runtime_version.py shown as missing for previous versions.
See [here](https://github.com/ansys/pydpf-core/actions/runs/11127183318) for the CI run with `protobuf<5.27.0`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "importlib-metadata >=4.0",
     "numpy<2",
     "packaging",
-    "protobuf",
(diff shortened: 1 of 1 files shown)


rafacanton (approved):

BClappe (approved):

PProfizi:
@rafacanton @BClappe after discussing with the PyAnsys team, it seems that we actually need to force a maximum version on the library used to build `ansys-grpd-dpf` as other pyansys libraries are not ready to bump to `protobuf>=5`

PR #1805, opened 2024-10-14
Fix tests.yml doctest run and coverage reporting

PProfizi:


PProfizi:
Hi @moe-ad I just realized I already had a PR to fix what I mentioned yesterday about artifact naming conflict.

moe-ad (approved):
LGTM!

PR #1821, opened 2024-10-21
Fix hidden operators exposure

PProfizi:
This PR completes a PR on the server-side to hide hidden operators.
It also refactors Python operator code generation for simplification and improved maintainability.


src/ansys/dpf/core/operators/build.py
```diff
@@ -182,13 +182,30 @@ def build_operators():
 
     succeeded = 0
     done = 0
+    hidden = 0
+    # List of hidden operators to still expose for retro-compatibility
+    # until they are fully deprecated
+    hidden_to_expose = [  # Use internal names
```
PProfizi, line 188: @cbellot000 here is the list of exceptions for hidden operators to still expose as a Python module

.ci/code_generation.py
```diff
@@ -65,41 +25,8 @@
...
-        print(f"Could not generate operators for optional library {lib}:\n{str(e)}")
-
-# Reorder imports alphabetically in __init__.py files to reduce changes raised
-for init_file_path in glob.glob(os.path.join(TARGET_PATH, "**/__init__.py"), recursive=True):
-    with open(init_file_path, "r") as init_file:
-        lines = init_file.readlines()
-    with open(init_file_path, "w") as init_file:
-        init_file.writelines(sorted(lines))
+core.set_default_server_context(core.AvailableServerContexts.premium)
+core.start_local_server(config=core.AvailableServerConfigs.LegacyGrpcServer)
```
cbellot000, line 30: why legacy??

cbellot000 (approved):

.ci/code_generation.py
PProfizi: @cbellot000 that is historical but we could try with the `GrpcServer`. In a next PR.

PR #1829, opened 2024-10-23
move contributing section into the getting started section

luisaFelixSalles:
Doc changes:

- moves contributing section into the getting started section
- navigation bar don't have "more" anymore

image


doc/source/conf.py
```diff
@@ -91,7 +91,8 @@
 
 redirects = {
      "user_guide/getting_started_with_dpf_server": "../getting_started/dpf_server.html",
-     "concepts/index": "../user_guide/index.html#concepts"
+     "concepts/index": "../user_guide/index.html#concepts",
+     "contributing": "source/getting_started/contributing.html"
```
PProfizi, line 95:
```suggestion
     "contributing": "getting_started/contributing.html"
```
Isn't that the correct new path?

doc/source/getting_started/index.rst
```diff
@@ -111,11 +111,14 @@ To use PyDPF-Core, in the same Python environment, run this command:
 
 
 .. toctree::
+   :maxdepth: 2
    :hidden:
+   :caption: Introduction
```
PProfizi, line 116: @luisaFelixSalles what does this render like? I think that no section title would be better for now. Until we have at least two sections.

PProfizi, line 116:
```suggestion
```

luisaFelixSalles: image

PR #1836, opened 2024-10-25
Add a vtk_mesh_is_valid helper to check validity of VTK meshes

PProfizi:
Add a 'check_validity' argument to the dpf_mesh_to_vtk function.

### DIFF
--- a/src/ansys/dpf/core/vtk_helper.py
+++ b/src/ansys/dpf/core/vtk_helper.py
@@ -19,8 +19,10 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import warnings
 
 import numpy as np
+import pyvista
 import pyvista as pv
 from typing import Union
 from vtk import (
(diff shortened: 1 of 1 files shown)


rafacanton (approved):

user27182:
> Hello @user27182, as long as the licensing terms at the beginning of the file containing the code of interest are respected I see no issue. 

Thanks for the feedback. I have started a draft PR for this functionality here: https://github.com/pyvista/pyvista/pull/8164
Although I used the code from this ansys PR as a starting point, the implementation is substantially different and has very little in common with the ansys code now. The only thing required by the ansys licensing terms that _would_ probably be an issue is the statement:
> "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."

since PyVista does not (yet) include these kinds of notices in the source code (though they probably should start), But again, I don't think substantial portions are included in the PyVista draft PR, so the licensing terms of the ansys code don't apply anyway.

> That would actually allow me to use this validation function directly from PyVista instead of maintaining it myself.

In the PyVista code from https://github.com/pyvista/pyvista/pull/8164,  the new `DatSet.validate_mesh` method is similar to the ansys `vtk_mesh_is_valid` function. Similarly, the PyVista function returns a `ValidationReport` dataclass, analogous to the ansys `VTKMeshValidity` dataclass.

The draft is mostly complete, and I _think_ the current API  should allow you to use the PyVista code directly to replicate the ansys code functionality in full.

But, it may be worthwhile for you to try out the PyVista API to see if there's anything missing that I should add.

PR #1839, opened 2024-10-28
any to and from workflow

cbellot000:


PProfizi (approved):

src/ansys/dpf/core/any.py
```diff
@@ -185,6 +186,11 @@ def _type_to_new_from_get_as_method(self, obj):
                 self._api.any_new_from_any_collection,
                 self._api.any_get_as_any_collection,
             )
+        elif issubclass(obj, workflow.Workflow):
+            return (
```
PProfizi, line 190: @cbellot000 is it normal for it to say it is not covered by testing?

PR #1859, opened 2024-11-04
Fix call to ansys/actions/doc-deploy-stable

PProfizi:


.github/workflows/releaser.yml
```diff
@@ -10,10 +10,6 @@ on:
         description: 'Release tag'
         default: 'latest'
         type: string
-      checkout_ref:
-        description: 'Ref to checkout for doc index (default to current branch)'
-        default: ''
-        type: string
```
BClappe, line 16: You don't need this anymore in any case? @PProfizi

BClappe (approved):

.github/workflows/releaser.yml
PProfizi: Yes, that was for the old version of the action using Meilisearch.

Issue #1831, opened 2024-10-23
Numpy 2.x compatibility


greschd:
What is the status of PyDPF w.r.t. compatibility with the `numpy` 2.x release? I have noticed that it is declared as `numpy<2`, but don't know whether there is a known incompatibility, or it is simply a conservative dependency declaration.

The reason I stumbled upon this is that the poetry dependency solver in PyACP (which does not restrict numpy):
- upgraded `numpy` to 2.x
- downgraded DPF core to `0.12.2`, which does not limit `numpy` (but presumably isn't any _more_ compatible)

By forcing `ansys-dpf-core>=0.13`, I could get it to pick the older numpy and newer DPF instead.

PR #1864, opened 2024-11-05
Remove numpy restriction to <2

PProfizi:
Starting with `numpy==2` there is an explicit distinction between `np.int32` and `np.int64` which results in a difference in string representations of certain numbers.
Solves #1831


RobPasMue (approved):
LGTM!

PProfizi:
> Isn't there any other modification to make anywhere else? Only 4 files?

Apparently, yes, I guess this is because there are only very specific places where we return an `np.int32`.

PR #1901, opened 2024-11-15
feat: add any_operator support

Matteo-Baussart-ANSYS:


src/ansys/dpf/core/dpf_operator.py
```diff
@@ -121,7 +121,7 @@ class Operator:
 
     """
 
-    def __init__(self, name, config=None, server=None):
+    def __init__(self, name=None, operator=None, config=None, server=None):
```
PProfizi, line 124: Hey @Matteo-Baussart-ANSYS, adding the new parameter in second position could be a break retro-compatibility for users who use positional arguments. I know many users who do not bother with using parameter names when calling functions.

PProfizi: Also, you can use `pre-commit` to have automatic styling when doing a commit, this way you will not get failing Style checks.

PProfizi: It also looks like this still requires changes server-side to be synchronized. FYI You can run tests and target a custom DPF server branch on `dpf-standalone` if you pushed one there.

Matteo-Baussart-ANSYS: Thank you for the feedback. I updated the `__init__` method and applied the style checks.
However, the [changes](https://tfs.ansys.com:8443/tfs/ANSYS_Development/DPF/_git/DPF/pullrequest/541803) required on dpf side are not merged yet. Should I wait for them to be merged, or can i specify a commit for testing ?

cbellot000 (approved):

PR #1927, opened 2024-11-25
Limit docstring testing in ci_release.yml

PProfizi:


ansjhajjar:
What's the reason for that ?

Issue #1409, opened 2024-02-09
Be able to read *.prs & *.psd files


jviqueg:
### Description of the feature

It would be useful for users to be able to get access to mode coefficients from *.prs in Response Spectrum and modal covariance matrices from *.psd in Random vibration analysis (see below), same way as we can now get access to Participation Factors from *.mode file coming from a modal analysis using dpf.Operator(r"spectrum_data").
![image](https://github.com/ansys/pydpf-core/assets/106327851/4c7e4f95-51ae-43c4-a7bc-e68bf9575564)
Thanks! :-)
FYI @cbellot000 & @ayush-kumar-423

oparreno:
@jviqueg, @ayush-kumar-423  starting in 25R2:
"spectrum_data" operator will allow to read participation_factors, mode_coefficients and damping_ratios from the PRS file. 
"psd_matrices_provider" operator will allow you to get the covariance matrices from a PSD file.

ayush-kumar-423:
Thanks @oparreno !

PProfizi:
@oparreno ok I just saw the changes in the last server sync PR, so it is now exposed in the master branch of PyDPF-Core.
Could you please describe what we need to add to the table of supported files [here](https://dpf.docs.pyansys.com/version/stable/index.html).

oparreno:
@PProfizi, 
For MAPDL file format, we should add .prs & .psd files starting in version Ansys 2025R2
![Image](https://github.com/user-attachments/assets/bb7f9dda-e498-4878-a05c-faf1259c9388)

PR #1937, opened 2024-11-26
Mark support for psd and prs files

PProfizi:
Following #1409


oparreno (approved):

PProfizi:
@oparreno do you think we could design an example?

PR #1940, opened 2024-11-27
Add a links and refs file to be used in the doc 

luisaFelixSalles:
Add a file with substitution text to be used around the doc for the main DPF objects and useful links


PProfizi (approved):
Great idea!

PR #1942, opened 2024-11-28
Remove exclude of vtk 9.4.0

PProfizi:
This PR tests and validates the bump of `pydpf-actions/prepare_tests` to using `pyvista/setup-headless-display-action@v3` which should solve the issue with `vtk==9.4.0`.
Note that `pyvista==0.44.2` was just released which also excludes `vtk==9.4.0` until everything is fixed on their side.


PProfizi:
Now a duplicate of #2229

PR #1946, opened 2024-11-29
Examples/refactor custom plugin examples

PProfizi:
Refactor examples on custom Python operators and plugins by creating helpers to download files.
This prepares their revamping into tutorials and hides the file retrieval logic from the user, which is something we do not need to show for the examples.

### DIFF
--- a/examples/08-python-operators/01-package_python_operators.py
+++ b/examples/08-python-operators/01-package_python_operators.py
@@ -54,35 +54,10 @@ For this example, the plug-in package contains two different operators:
 #
 # Download the ``average_filter_plugin`` plug-in package that has already been
 # created for you.
-
-import os
-
 from ansys.dpf.core import examples
 
-
-print("\033[1m average_filter_plugin")
-file_list = [
-    "average_filter_plugin/__init__.py",
-    "average_filter_plugin/operators.py",
-    "average_filter_plugin/operators_loader.py",
-    "average_filter_plugin/common.py",
-]
-plugin_folder = None
-GITHUB_SOURCE_URL = (
-    "https://github.com/ansys/pydpf-core/raw/"
-    "master/doc/source/examples/07-python-operators/plugins/"
-)
-
-for file in file_list:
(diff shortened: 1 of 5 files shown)


examples/08-python-operators/01-package_python_operators.py
```diff
@@ -54,35 +54,10 @@
...
-    operator_file_path = examples.downloads._retrieve_file(EXAMPLE_FILE, file, "python_plugins")
-    plugin_folder = os.path.dirname(operator_file_path)
-    print(f"\033[1m {file}:\n \033[0m")
-    with open(operator_file_path, "r") as f:
-        for line in f.readlines():
-            print("\t\t\t" + line)
-    print("\n\n")
-
+plugin_folder = examples.download_average_filter_plugin()
+print(plugin_folder)
```
PProfizi, line 60:
```suggestion
```

src/ansys/dpf/core/examples/downloads.py
```diff
@@ -1993,3 +2000,142 @@ def find_distributed_msup_folder(
...
+    EXAMPLE_FILE = GITHUB_SOURCE_URL + file_name
+    operator_file_path = _retrieve_file(
+        EXAMPLE_FILE, filename=file_name, directory="python_plugins"
+    )
+    return find_files(operator_file_path, should_upload, server, return_local_path)
+
+
+def _retrieve_plugin(
+    file_list: list[str], should_upload: bool = True, server=None, return_local_path=False
+) -> Union[str, None]:
```
ansjhajjar, line 2133: This can also be seen as a `Optional[str]`
But in the same time python3.10 expects us to write it as a `str | None` so the union might be more future-proof

ansjhajjar (approved):

src/ansys/dpf/core/examples/downloads.py
PProfizi: Until we drop support for 3.9 completely I tend to use `Union`.

PR #1952, opened 2024-12-04
doc: use sphinx-autoapi for API reference

moe-ad:
This PR is related to ansys-internal/pyansys-planning#100. The migration of the API documentation migration is being handled here.

### DIFF
--- a/doc/source/api/index.rst
+++ b/doc/source/api/index.rst
@@ -1,10 +1,11 @@
-.. _ref_api_section:
-
-=============
 API reference
 =============
+
+This section describes PyDPF-Core endpoints, their capabilities, and how
+to interact with them programmatically.
+
 .. toctree::
-   :maxdepth: 2
-   :caption: API reference
-   
-   ansys.dpf.core
+   :titlesonly:
+   :maxdepth: 3
+
+   <span class="nf nf-md-package"></span> ansys.dpf.core</api/ansys/dpf/core/index>
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -62,7 +62,31 @@ for example in glob(r"../../examples/**/*.py"):
         ignored_pattern += f"|{example_name}"
 ignored_pattern += "|11-server_types.py"
 ignored_pattern += "|06-distributed_stress_averaging.py"
+ignored_pattern += "|02-python_operators_with_dependencies.py"
+ignored_pattern += "|00-fluids_model.py"
 ignored_pattern += r")"
+
+# Autoapi ignore pattern
+autoapi_ignore_list = [
+    "*/log.py",
+    "*/help.py",
+    "*/mapping_types.py",
+    "*/ipconfig.py",
+    "*/field_base.py",
+    "*/cache.py",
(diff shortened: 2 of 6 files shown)


moe-ad:
@jorgepiloto, something I need to mention (already discussed with Revathy about this too).
If you try to build with sphinx 7.1.0 (which is the current default on this repo) while having autoapi enabled, the build fails. We narrowed it down to **app.confdir** being a str object in sphinx 7.1.0 and a path object in sphinx 8.1.3.

In ansys-sphinx-theme 1.2.2 (which is the current default on this repo), the autoapi extension treats app.confdir correctly (as a path object) but using sphinx 7.1.0 will result in an error because of the above reason.

In order words, we need sphinx 8.1.3 to be enforced moving forward.

jorgepiloto:
Thanks for the research, @moe-ad. Then, let's focus on updating the rules for the moment. After that, we can update the examples to Jupytext. This should allow us to revisit the complete set of examples, making sure they run without random failures.

moe-ad:
Finally... :(

doc/source/conf.py
```diff
@@ -64,6 +64,28 @@
 ignored_pattern += "|06-distributed_stress_averaging.py"
 ignored_pattern += r")"
 
+# Autoapi ignore pattern
```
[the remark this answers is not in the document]
moe-ad: The files are excluded in the current apidoc generation (see [the original file](https://github.com/ansys/pydpf-core/blob/6f7cde8ac0b85b86c6173246a91cd49846c7500a/.ci/build_doc.bat#L2)). So I did the same thing when using autoapi.
From my observation, the reason why those files are being ignored originally is that they are meant for internal use within the package and not meant to be exposed to users for usage

doc/source/conf.py
```diff
@@ -217,9 +234,35 @@ def reset_servers(gallery_conf, fname, when):
...
+        "ignore": autoapi_ignore_list,
+        "add_toctree_entry": True,
+        "member_order": "bysource",
+    }
 }
 
-
+# Configuration for Sphinx autoapi
+suppress_warnings = [
+    "autoapi.python_import_resolution",
```
jorgepiloto, line 261: Why do we have to suppress this warning?

jorgepiloto: Oh! I see... It is because we may be hitting cyclic imports. Could you add a TODO in here @moe-ad and open an issue so we do not forget about removing this in the future once the code is healthier?

jorgepiloto (approved):

requirements/requirements_docs.txt
```diff
@@ -1,4 +1,5 @@
 ansys-sphinx-theme==1.2.2
```
jorgepiloto, line 1:
```suggestion
```

doc/source/api/index.rst
```diff
@@ -1,10 +1,11 @@
-.. _ref_api_section:
```
PProfizi, line 1: Hi @moe-ad, does this not break any links in the doc?

Issue #296, opened 2022-07-08
Update package layout to PyAnsys guidelines


jorgepiloto:
As with other projects in the PyAnsys ecosystem, the layout of this project should adapt to the general guidelines.

## Low effort

- [ ] Enable Ruff rules
  - [ ] PyCodestyle
  - [x] PyDocstyle, see https://github.com/ansys/pydpf-core/pull/1975
  - [ ] PyFlakes, see https://github.com/ansys/pydpf-core/pull/1953
  - [x] Isort, see #2021 
  - [ ] PEP8-Naming
  - [x] Flake8-use-pathlib, see https://github.com/ansys/pydpf-core/pull/1949
  - [x] Flake8-todos, see https://github.com/ansys/pydpf-core/pull/1981

- [ ] Remove autoapi warning suppression once the codebase is healthier. See #1967 
- [x] Address removed reference labels from module docstrings. See #1980 

## Intermediate effort

- [x] Use a `src/` layout. Just need to move the "ansys/" dir inside a "src/" one and fix some links to this new location: #664

- [x] Use a `pyproject.toml` and drop any `setup.py` or `setup.cfg` files. https://github.com/ansys/pydpf-core/pull/909

- [x] Rename ``/docs`` directory to ``/doc``: Link issue here.

- [x] Implement Tox for the project
  - [x] Drop any patch scripts in favor of Tox 
    - #1989 
    - #2015 
    - #2031
    - #2056

- [x] Use sphinx-autoapi for generating the API reference, see https://github.com/ansys/pydpf-core/pull/1952
- [x] Improve static search live preview performance and fix missing module icons in the api documentation, see #1971 

## High effort

- [ ] Use containers from the GitHub Container Registry in the CI/CD
- [ ] Use semver tags

More about packaging guidelines in https://dev.docs.pyansys.com/packaging/structure.html.

jorgepiloto:
@pyansys-automation never close issue

jorgepiloto:
This issue is a meta-issue. Efforts for implementing and fixing its child issues have been made over the last years. Thus, we'll close it once all the minor issues have been fixed.

PR #1953, opened 2024-12-04
fix: pyflakes style

jorgepiloto:
This is related with https://github.com/ansys/pydpf-core/issues/296. It enables the PyFlakes (F) rule.


jorgepiloto:
Checking the decrease in coverage. Some tests were duplicated and exactly the same.

tests/test_field.py
jorgepiloto, line 1: This test slightly differs from another named exactly the same within this file. To be reviewed...

jorgepiloto:
Progress on this for the next days:

- Restore unused properties
- Verify their value in Pytest

This solves the issues for coverage and codacy.

Example:

```python
mesh = Object.mesh
assert np.assert_allclose(mesh, mesh_expected, atol, rtol)
```

moe-ad:
I do not see how we can boost coverage on the erring lines. Most are modified f-strings or exception statements. If the lines are really being covered before now, there is no reason why they shouldn't be covered after the changes, at least I can't think of any.

moe-ad:
@jorgepiloto , just a reminder that I think we can merge this?

PProfizi:
Hi @jorgepiloto, hi @moe-ad is this still active? Should it be marked as ready-for-review?

PR #1949, opened 2024-12-03
maint: modifying files to conform with "PTH" ruff rule

moe-ad:
This PR is associated with #296. Flake9-use-pathlib ruff rule will be addressed via this PR.

PR #1956, opened 2024-12-05
Junk: will delete the branch shortly

moe-ad:
See title

### DIFF
--- a/.ci/build_wheel.py
+++ b/.ci/build_wheel.py
@@ -3,6 +3,7 @@
 
 import argparse
 import subprocess
+from pathlib import Path
 import os
 import sys
 import shutil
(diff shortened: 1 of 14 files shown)


moe-ad:
Now let's start adding changes incrementally.

moe-ad:
Looks like files in .ci/ are not the culprits.

moe-ad:
So one issue discovered. Tests are being made with py3.9 which doesn't support Path.walk(). Causing one failure at this stage.

moe-ad:
It's probably better to work with py3.9 as I do not yet know the full implications of switching the CI/CD test environments to py3.12. I can see a different test now failing when the switch is made to py3.12 in the last commit.

moe-ad:
Changes to the example files causes no test issues.

moe-ad:
That works. Let's try another one.

moe-ad:
mentioning #1949.

PR #1960, opened 2024-12-11
fix: exclusion of example files causing doc-build failure

moe-ad:
Certain examples are causing the documentation build process to fail and this is blocking the merging of several other pull requests (see #1952, #1959). It is proposed to temporarily exclude erring examples from documentation generation until there behavior is fixed.


PProfizi:
Hi @moe-ad,
can I have more information on how/why this example fails before ignoring it?

PProfizi:
Hi @moe-ad closing as not required anymore

PR #1961, opened 2024-12-11
Add contact result properties

janvonrickenbach:
This is required so that force_elemental_nodal is correctly computed for contact results in pydpf-post. As far as I can see this dictionary is only used to lookup metadata on request and does not impact the available results directly. The scripting name is taken from dataProcessingDoc.html in the dpf core repo.


cbellot000 (approved):

PProfizi:
Hi @cbellot000 should this dictionary even exist? Shouldn't that information be given by the server?

PR #1965, opened 2024-12-12
feat: exposing quantity types in PyDPF Core

a-bouth:
Exposing the `quantity_type`for `FieldsDefinition`

# Main Changes

* Added calls to the C API in `field_definition_capi.py`
* Adding methods in `field_definition.py`
* * Note: I initially wanted to use a `@property` decorator for `quantity_type`, but it turns out that the getter needs an index, which we cannot specify with the `@property` decorator. I opted for defining explicits `get_quantity_type(index)` and `set_quantity_type()` methods. Please let me know if I should do it differently.
* * Also exposed `is_of_quantity_type()` and `get_num_quantity_types_available()`
* Added corresponding test in `test_field.py` (I found other tests for `FieldDefinition` here, but maybe that's not the best place ?)

# Side changes

Enabling the cast of `FieldsContainer` in the `Any` class


src/ansys/dpf/gate/generated/field_definition_capi.py
```diff
@@ -184,6 +184,35 @@ def csfield_definition_get_shell_layers(res):
 		if errorSize.value != 0:
 			raise errors.DPFServerException(sError.value)
 		return res
+	
+	@staticmethod
```
cbellot000, line 188: this is supposed to be generated, how did you do it?

src/ansys/dpf/core/field_definition.py
```diff
@@ -153,6 +153,61 @@ def dimensionality(self):
         self._api.csfield_definition_fill_dimensionality(self, dim, nature, dim.internal_size)
         return Dimensionality(dim.tolist(), natures(int(nature)))
 
+    def get_quantity_type(self, index=0):
```
cbellot000, line 156: Maybe having just one property which returns the quantity_types would be easier too use, what do you think?

tests/test_field.py
```diff
@@ -517,6 +517,37 @@ def test_create_and_update_field_definition(server_type):
     assert fieldDef.location == locations.nodal
 
 
+def test_field_definition_quantity_type(server_type):
+    fieldDef = FieldDefinition(server=server_type)
+
+    # Testing the setter
+    qt = "my_quantity_type"
+    fieldDef.set_quantity_type(qt)
```
cbellot000, line 525: is it rather an add_entity_type?

src/ansys/dpf/core/field_definition.py
```diff
@@ -153,6 +153,61 @@ def dimensionality(self):
...
+        """Setter for Quantity Type
+
+        Parameters
+        ----------
+        quantity_type: Quantity Type
+            Quantity type to set
+        """
+        self._api.csfield_definition_set_quantity_type(self, quantity_type)
+
+    def get_num_available_quantity_types(self):
```
cbellot000, line 184: to be more pythonic, maybe we could have a property "num_quantity_types" ?

cbellot000 (approved):

tests/test_field.py
a-bouth: @cbellot000 

I used that name in order to be consistent with the C API that uses :

``` C++
ansys::dpf::FieldDefinition fd;
fd.setQuantityType()
```
But I agree that **add**_quantity_type would make more sense than set_quantity_type

... but - regarding one of your other comment where you suggest to have just one property for quantity_type - I would say that we should not define a setter using the decorator for quantity_type otherwise it would be confusing (because the setter would actually _add_ a quantity type, which is counter intuitive)

What I'm saying: 

We should have 

``` python
@property 
def quantity_type(self): 
   # Returns all quantity_types and makes quantity_type as a property

def add_quantity_type(self, quantity_type):
   # Add a new quantity_type
```

But we should NOT have:

``` python
@quantity_type.setter
def quantity_type(self,qt): 
   # adds a quantity type
```

Would that be fine with you ?

src/ansys/dpf/gate/generated/field_definition_capi.py
a-bouth: @cbellot000  this is fixed, I used the generated code and removed the added one

src/ansys/dpf/core/field_definition.py
a-bouth: @cbellot000 done

cbellot000 (approved):

cbellot000:
[like]  Camille Bellot reacted to your message:
________________________________

PR #1971, opened 2024-12-18
Doc: improve static search speed and fix missing module icons

### LINKED ISSUE #1970: Improve static search live preview performance and fix missing module icons in the api documentation
After the recent migration to autoapi (see #1952), it was observed that module icons are missing in the section navigation:
Image
Also, the static search live search preview was observed to be slow.
This issue will address both concerns.

moe-ad:
This PR will close #1970. It adds html theme option to improve static search live preview performance. It also fixes the problem of missing module icons in the section navigation of the api documentation.


jorgepiloto (approved):
Excellent, @moe-ad. I hope you used a script to automate this.

moe-ad:
Yes. I used a python script for the task.

src/ansys/dpf/core/operators/averaging/elemental_difference.py
```diff
@@ -1,6 +1,6 @@
 """
 elemental_difference
-====================
+
```
PProfizi, line 3: Hi @moe-ad, everything under operators/ is generated automatically using a template. We'll need to update the template for this to stay that way.

PProfizi: See [here](https://github.com/ansys/pydpf-core/blob/b7f89f7015c15746b8b95c3a2d7f7ac855d91359/src/ansys/dpf/core/operators/operator.mustache#L3)

PR #1975, opened 2024-12-23
doc: pydocstyle ruff rule

moe-ad:
This PR is related to #296. The codebase will be modified to conform with pydocstyle guidelines.


pyproject.toml
jorgepiloto, line 1: How are we excluding the auto-generated files from being linted?

jorgepiloto (approved):
Amazing work, @moe-ad. Just left one question.

pyproject.toml
moe-ad: There is a [regular expression](https://github.com/ansys/pydpf-core/blob/0d7bba792e11198cce47bcaabd7075b648be111c/.pre-commit-config.yaml#L1) in pre-commit-config.yaml that matching directories we are excluding.

moe-ad:
Thank you @jorgepiloto... 🕺

pyproject.toml
jorgepiloto: Oh! Thanks for clarifying. Then, let's get this merged.

PR #1981, opened 2025-01-02
Doc: flake8 todos

moe-ad:
Related to #296. flake8-todos ruff rule is enabled via this PR


RobPasMue:
Hi @moe-ad! Could we separate this PR in two and have the license headers updated in a different PR? Otherwise it makes it difficult to review...

RobPasMue:
Also, the TODO URL should point to an issue where we are keeping track of that deprecation. We shouldn't be pointing to a PR URL. See https://github.com/ansys/pyansys-geometry/issues/1319 as an example. The first time I implemented the flake8-todo task, I created this longterm tracker. Ideally, each todo should have an independent issue though.

RobPasMue:
IMO the only way you will actually keep track of work is through issues, not PRs. It's hard to keep track of them otherwise. I'd advise to create a longterm tracker and let the @ansys/pydpf-admins dev team handle them whenever they have the chance.

moe-ad:
@RobPasMue thanks!
I have implemented your suggestion.

RobPasMue (approved):
LGTM!

PR #1985, opened 2025-01-02
Remove class title underline from operator template

PProfizi:
Following changes made to fix problems in the API documentation.

As mentioned [here](https://github.com/ansys/pydpf-core/pull/1971#discussion_r1900968618), changes were made directly to the operator modules despite them being generated automatically upon server sync. This currently creates merge conflicts in the server sync pipeline.
This PR removes the conflict by updating the template used when generating the modules so they match with the expected format.


PProfizi:
@moe-ad @jorgepiloto how should I mark the operators modules as not to be edited directly? 
As can mark them as requiring approval from maintainers as it is done for `src/ansys/dpf/gate/` or `src/ansys/grpc/dpf` yet that is not exactly the same thing.

moe-ad:
Hello @PProfizi. Welcome back! Thanks for identifying this.
Regarding your question, we can add a pre-commit hook that will not allow commits containing changes to that directory by any author/committer other than the one configured for the CI action that automatically generates those files.
Maybe @jorgepiloto can also have a different suggestion.

moe-ad (approved):
LGTM!

PR #1988, opened 2025-01-03
Fix averaging example plots

PProfizi:
The first example of the averaging section lost its calls to plot methods at some point, not fitting the description anymore. This PR adds them back to fix the example.

### DIFF
--- a/examples/09-averaging/00-compute_and_average.py
+++ b/examples/09-averaging/00-compute_and_average.py
@@ -120,7 +120,6 @@ def compute_von_mises_then_average(analysis):
     max_val = min_max.outputs.field_max()
 
     avg_von_mises.plot()
-  
     return max_val.data[0]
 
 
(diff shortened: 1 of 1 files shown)


PProfizi:
Hi @moe-ad, your PR #1953 made me realize there is an issue with the example where plots were no longer generated.

examples/09-averaging/00-compute_and_average.py
```diff
@@ -120,6 +119,8 @@ def compute_von_mises_then_average(analysis):
     min_max.inputs.field.connect(avg_von_mises)
     max_val = min_max.outputs.field_max()
 
+    avg_von_mises.plot()
+
```
PProfizi, line 123:
```suggestion

```

examples/09-averaging/00-compute_and_average.py
```diff
@@ -157,6 +157,8 @@ def average_then_compute_von_mises(analysis):
...
```
PProfizi, line 161:
```suggestion
```

PR #1989, opened 2025-01-06
Feat: tox implementation

moe-ad:
This PR is related to #296.
This is not a catch all PR, the idea is to first migrate some CI workflows to tox and gradually extend to other CI workflows via other PRs.


moe-ad:
PyDPF uses a custom wheel building step in CI. The first commit implements a way of integrating that step with tox.

moe-ad:
@PProfizi 
Now simply running locally:
- `tox -e runtests-parallel` should run all tests in parallel
-  `tox -e runtests-sequential` should run all tests sequentially
With automated test target organization and server process(es) shutdown before and after all tests have run.

Per test (irrespective of the running mode), a fresh environment with individual `ansys.dpf.core` package installation is used for test runs. The sacrifice with running in parallel is not having detailed output to `stdout`, only success/failure summary is shown for each test environment. Possible to force detailed output to `stdout` but that will result in an unstructured report. Maybe there is some log file being generated that can be inspected at the end of all runs?

Next steps:
- Confirm all environment variables we need to be passing to the test environment since tox won't automatically pass them.
- The package building step is done for each test environment. Will be nice if it can be done just once, and then installed into each environment as needed. Tox is smart enough to do this if package building were via a pyproject.toml or something similar.

tox.ini
```diff
@@ -0,0 +1,125 @@
...
+    bash -c '\
+    if [ {on_platform} == "linux" ]; then \
+        export platform="manylinux_2_17"; \
+    elif [ {on_platform} == "win32" ]; then \
+        export platform="win"; \
+    else \
+        echo "Unknown OS"; \
+    fi; \
+    echo $platform; \
+    python .ci/build_wheel.py -p $platform -w'
```
jorgepiloto, line 44: Tox is able to detect the platform it is running. Just use the following:

```ini
platform =
    linux: linux
    windows: win32
```

From then one, you can execute commands based on this platform by using this form:

```ini
commands =
    linux: ls
    windows: dir
```

jorgepiloto (changes requested):
Thanks for working on this, @moe-ad. I know it is still in draft mode, but wanted to raise a point about auto-detecting the platform.

tox.ini
moe-ad: @jorgepiloto I have been thinking about this for a while. I have had another go at it after reading your comments without success.

It won't work via the above suggestions, at least based on what I have tested. tox expects any factor being used for conditional settings to be present in the environment name itself, otherwise the commands linked to the linux/windows factors will never be executed when tox encounters them. At the same time, the naming specification of an external wheel building environment must have the format [<package_env>_external](https://tox.wiki/en/latest/config.html#external-package-builder), meaning we can't rename the wheel building environment to `[<package_env>-{linux,windows}_external]` or something similar, and because of that the commands are not executed when tox encounters them. Maybe there is something I am missing? 

All of these complications can be significantly reduced if there is a way to specify the wheel building logic directly within pyproject.toml. Being able to do this will also simplify some other things. @PProfizi was asking if this is possible yesterday, any ideas?

moe-ad:
@jorgepiloto thanks for our discussion about this, I have improved the configuration significantly by taking advantage of the `depends` feature in tox in a way that makes test invocation more intuitive:
- `tox` / `tox --parallel` will run all tests in sequential/parallel mode.
- `tox -e pretest,<test-list>,posttest` will run specific tests. This invocation can also be used with the parallel flag.

@PProfizi we will merge this PR as Jorge suggested that we should make this available for users to start taking advantage of tox locally immediately. We will proceed with next steps in other PRs.

tox.ini
jorgepiloto: From our private discussion, it looks like this is necessary unless we move the `build_external` logic to the `setup.py`, which is the standard approach for dynamic installations.

Also, is this cross-platform? Have you tried this in a Windows platform? I do not think the "bash" command will work there.

moe-ad: Thanks for catching that. I had git bash installed on windows which allowed it to work cross platform, might not be the case for other users truly. Implemented your suggestion by taking advantage of `{on_platform}` tox substitution and accounting for that in wheel building script, let me know if this approach is okay.

**EXAMPLES**
Running **`tox -e pretest,test-launcher,posttest`** on windows:
image


Running **`tox -e pretest,test-launcher,posttest`** on linux:
image

tox.ini
```diff
@@ -0,0 +1,121 @@
...
+    ANSYSLMD_LICENSE_FILE
+    AWP_ROOT242
+
+package = external # To allow custom wheel builds
+
+[testenv:build_external]
+description = Environment for custom build of package wheels, solves PyDPF custom wheel building requirement
+
+allowlist_externals =
+    bash
```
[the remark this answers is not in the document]
moe-ad: Removed. Thanks for catching that.

PR #2001, opened 2025-01-10
Accept a Python range when setting scoping IDs.

PProfizi:
Hi @JennaPaikowsky, the PR for the field inspired me to do this for the `Scoping` class as I noticed we were doing `list(range(*))` in the docstring examples for the `Field` when defining a `Scoping`.

This PR mainly adds the support for a range as input when setting the IDs of a scoping.

It also adds type hinting and improves the docstrings a bit. I'll also add some docstring examples and testing.

I'll actually add the `AnyServerType` type for type hint in a dedicated PR if @moe-ad and @jorgepiloto confirm this is the correct way to provide a combination of allowed types for type hints in the rest of the library. See #2002 

@jorgepiloto @moe-ad one thing I would like our opinion on is the type hint for pointers and Grpc messages as I think it will not work as-is. Edit: it looks like it does! Even though the IDE is complaining that `ansys.grpc.dpf.scoping_pb2.Scoping` does not exist.

### DIFF
--- a/src/ansys/dpf/core/scoping.py
+++ b/src/ansys/dpf/core/scoping.py
@@ -43,7 +43,6 @@ from ansys.dpf.gate import (
     dpf_vector_capi,
     dpf_vector_abstract_api,
     dpf_vector,
-    dpf_array,
     utils,
 )
 from ansys.dpf.gate.dpf_array import DPFArray
@@ -232,10 +231,26 @@ class Scoping:
             self._api.scoping_get_ids_for_dpf_vector(
                 self, vec, vec.internal_data, vec.internal_size
             )
-            return dpf_array.DPFArray(vec) if np_array else vec.np_array.tolist()
+            return DPFArray(vec) if np_array else vec.np_array.tolist()
 
         except NotImplementedError:
             return self._api.scoping_get_ids(self, np_array)
+
+    def get_ids(self, np_array: bool = False) -> Union[list[int], DPFArray]:
+        """Return the list of entity IDs in the scoping as a list or as a numpy.ndarray.
+
+        Parameters
+        ----------
+        np_array:
+            Whether to return the list of IDs as a numpy array.
+
+        Returns
+        -------
+        ids:
+            List of entity IDs in the scoping.
+
+        """
+        return self._get_ids(np_array=np_array)
 
     def set_id(self, index: int, scopingid: int):
         """Set the ID of the entity at index in the scoping.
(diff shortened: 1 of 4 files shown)


rafacanton:
@PProfizi I see post ci failing, do we need a PR there to support something related to the server? Otherwise, this LGTM!

PProfizi:
Ah, indeed, @rafacanton the scoping factory for named selections has a `server` parameter which is not used. I removed it but of course this is not backwards compatible with existing calls.

src/ansys/dpf/core/scoping.py
```diff
@@ -251,73 +278,79 @@ def _get_index(self, scopingid):
...
+    def ids(self, value: IdVectorType):
         self._set_ids(value)
 
     @property
-    def location(self):
-        """Location of the IDs as a string, such as ``"nodal"``, ``"elemental"``, and ``"time_freq"``.
+    def location(self) -> str:
+        """Location of the scoping as a string.
+
+        This defines the type of entity the IDs stored correspond to.
```
JennaPaikowsky, line 341: I am having a bit of trouble parsing this. Is it defining the type of entity stored in an ID? Defining an entity corresponding to a stored ID?

src/ansys/dpf/core/scoping.py
```diff
@@ -362,12 +395,13 @@ def __setitem__(self, index, id):
...
     @property
-    def size(self):
+    def size(self) -> int:
         """Length of the list of IDs.
 
         Returns
         -------
-        size : int
+        size:
+            Size if the scoping.
```
JennaPaikowsky, line 404:
```suggestion
            Size of the scoping.
```

JennaPaikowsky (approved):
A couple of minor points.

src/ansys/dpf/core/scoping.py
PProfizi: @JennaPaikowsky this defines what type the entities are. The entity IDs can be node IDs, element IDs, face IDs, etc...
Not sure how to formulate that properly.

JennaPaikowsky: Ah, got it.

"This defines what type of entity the IDs correspond to (such as node ID, element ID, face ID, and so on)."

PProfizi:
```suggestion
        This defines the type of entity the IDs correspond to (such as node ID, element ID, face ID, and so on).
```

PR #2012, opened 2025-01-13
feat: change tox environments logic

### LINKED ISSUE #2010: Move custom wheel building step in tox.ini
It has been decided that custom wheel building is not required for testing environments. In this issue:
- A separate environment should be created for custom wheel building.
- The test environments should simply make use of the default tox package building and installation.

moe-ad:
This will close #2010.


jorgepiloto (approved):
Great. Thanks Mo!

PR #2013, opened 2025-01-13
Fix server_to_ansys_version

PProfizi:
Realized the current implementation of the `ServerToAnsysVersion` map would not give the right answer for `10.0` (it gives `10R0`).
Fixed this and refactored to not need to update this version map anymore for each new version.

The only requirement is for DPF to continue only bumping major for new Ansys releases.

### DIFF
--- a/src/ansys/dpf/core/_version.py
+++ b/src/ansys/dpf/core/_version.py
@@ -31,18 +31,14 @@ min_server_version = "4.0"
 class ServerToAnsysVersion:
     def __getitem__(self, item):
         version = parse_version(item)
-        if version.major < 2024:
-            # Support the current DPF versioning scheme (XX.Y where necessarily XX<2024)
-            # Compute release version equivalent (YEAR+'R'+REVISION)
-            # The revision is 'R1' for any odd major DPF version, 'R2' for even major versions.
-            ansys_revision = 2 - version.major % 2
-            # The year is 2021 for DPF 1.0, and bumped every two releases.
-            ansys_year = 2020 + version.major // 2 + version.major % 2
-            # Return the corresponding Ansys release
-            return f"{ansys_year}R{ansys_revision}"
-        else:
-            # Support the YEAR.REVISION versioning scheme for DPF
-            return f"{version.major}R{version.minor}"
+        # The current DPF versioning scheme is MAJOR.MINOR.PATCH
+        # Compute release version equivalent (YEAR+'R'+REVISION)
+        # The revision is 'R1' for any odd major DPF version, 'R2' for even major versions.
+        ansys_revision = 2 - version.major % 2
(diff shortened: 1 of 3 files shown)


tests/test_version.py
```diff
@@ -0,0 +1,43 @@
...
+
+@pytest.mark.parametrize(
+    "server_version,ansys_version",
+    [
+        ("1.0", "2021R1"),
+        ("2.0", "2021R2"),
+        ("2.1", "2021R2"),
+        ("3.0", "2022R1"),
+        ("2023.0", "3032R1"),
+        ("2023.1.12", "3032R1"),
```
BClappe, line 36: 3032R1? Is it what you really wanted to do?

src/ansys/dpf/core/_version.py
```diff
@@ -27,31 +27,16 @@
...
-
     def __getitem__(self, item):
-        if len(item) == 3:
-            return self.legacy_version_map[item]
-        else:
-            split = item.split(".")
-            return split[0] + "R" + split[1]
+        split = item.split(".")
+        major = int(split[0])
+        minor = int(split[1])
```
BClappe, line 33: So now you are passing "2021.1", "2024.2", "2025.2" etc... Right ?

src/ansys/dpf/core/_version.py
```diff
@@ -27,31 +27,16 @@
...
+        if major < 2024:
+            # Detect new cases of XX.Y versions (necessarily XX<2024)
+            # Compute release version equivalent
+            minor = 2 - major % 2
+            major = 2020 + major // 2 + major % 2
```
BClappe, line 38: I don't understand the logic behind it sorry, could you provide some more information?

tests/test_version.py
PProfizi: Yes, in this case `2023.1.12` is what we would end-up for Ansys 3032R1 if we continue with the current versioning logic for DPF.

src/ansys/dpf/core/_version.py
PProfizi: Well we always get a version defined by a string with dots as separators.
The current versioning is at `10.0` and major is bumped for every Ansys release.
I can actually use the versioning library here to parse.
We are not currently using a versioning pattern such as `YEAR.REVISION`.
All I did was add support for it.

src/ansys/dpf/core/_version.py
PProfizi: Here this recreates the map between DPF version and Ansys version.
I'll add comments.

BClappe (approved):

rafacanton (approved):

guptadivya1501 (approved):

PR #2015, opened 2025-01-14
feat: add documentation build environment to tox.ini

### LINKED ISSUE #2011: Add doc build environment to tox.ini
As the title says

moe-ad:
As the title states. This closes #2011.


tox.ini
```diff
@@ -115,4 +115,48 @@ deps =
     -r requirements/requirements_test.txt
 
 commands =
-    pytest {env:PYTEST_PYTHON_FILES} {env:DEBUG} {env:COVERAGE} {env:RERUNS} {env:JUNITXML}
\ No newline at end of file
+    pytest {env:PYTEST_PYTHON_FILES} {env:DEBUG} {env:COVERAGE} {env:RERUNS} {env:JUNITXML}
+
+[testenv:doc-html]
```
jorgepiloto, line 120: Any chance we can have a `clean` and a `links` environment for the doc?

jorgepiloto (approved):
Awesome, @moe-ad. Would it be possible to have the `commands_post` at `[testenv]` level? That would apply it for every environment, right?

moe-ad:
@jorgepiloto Yes, indeed, that is possible at the `[testenv] `level. It will affect all environments and in environments where the commands are not needed, we can override `command_pre/post` to do nothing in those.

One problem is when you want to run some environments in parallel, the nature of what we are doing in commands_pre/post for PyDPF means that environments that finish executing earlier will try to kill the processes of currently running environments.

Another problem even in a sequential run is this:
- This workflow is not possible: `commands_pre` -> `test_a` -> `test_b` -> `commands_post`
- What will instead happen is: `commands_pre` -> `test_a` -> `commands_post` -> `commands_pre` -> `test_b` -> `commands_post`

And the first workflow is desired in the case of PyDPF.

jorgepiloto:
Thanks for the detailed explanation, @moe-ad. Then, feel free to merge when ready.

PR #2021, opened 2025-01-20
feat: enable isort ruff rule

moe-ad:
As the title describes.


moe-ad:
@jorgepiloto @PProfizi 
The documentation renders correctly as Jorge already predicted.

jorgepiloto (approved):

PR #2025, opened 2025-01-20
Accept rank-1 NumPy arrays for workflow and operator inputs which accept lists 

cbellot000:
This adds support for giving lists to operator or workflow inputs in the format of a rank 1 NumPy array.


PProfizi (changes requested):
Not required anymore, and raises Code Quality errors

PR #2031, opened 2025-01-22
ci: use tox in ci

moe-ad:
Related to #296


moe-ad:
Moving to "Ready for review" so I can observe workflow runs.

moe-ad:
Tests run fine now but overall takes two time as long as current CI.
Maybe we can test tox-uv out here @jorgepiloto? I can see much of the delay is due to the time it takes tox to spin up a new environment for each test target.

But at least we are now benefitting from the fact that each testing environment is clearly isolated, the random test failures should occur sparingly in this case.

jorgepiloto:
This is great news, @moe-ad. Let's give it a try to ' tox-uv' plugin, in that case. Just to make sure the bottle neck is now Tox installing all the necessary dependencies for every environment.

moe-ad:
pytest-xdist seems problematic in retro versions. May have to remove it for those.
> 15 mins being saved by using uv from my observation, but still not low enough.
One other thing to try is activating caching.

moe-ad:
More summary of my efforts so far:
- For testing within tox environments, vtk==9.3.1 doesn't work on Linux but works fine on Windows. However, vtk-osmesa==9.2.20230527.dev0 works fine across board. So the solution was to use the latter for Linux.
- Running tests locally on Linux through WSL works fine by taking note of the above @PProfizi.
- By combining `tox-uv`, `pytest-xdist`, and `tox run-parallel` in the different forms, the shortest CI test time achievable has been around 40 mins.
- However, running things in parallel (either via `tox run-parallel `or `pytest-xdist`) is not stable. This is because tests step on each other's toes (e.g., tests in different subprocesses trying to access the same file), leading to random "race conditions" and "segmentation faults".
- When the above occurs is random. For example, consider the case where _Test A_ (in _subprocess A_) and _Test B_ (in _subprocess B_) are configured to access the same file.
  - In one workflow run, _Test A_ may have finished execution before _Test B_ accesses the same file. In this case, no problem will occur.
  - In another workflow run, they may be trying to access the same file at the same time, and this will result in an error.
  - This results in random workflow failures each time the CI is run. For instance, the workflows that fail in the picture below may pass when the CI is run again, but a different set of workflows will then fail:
image

- This means that for us to take advantage of parallelization, the test suite needs some refactoring.

jorgepiloto (approved):
This is super, @moe-ad! Thanks for adding all these in place.

moe-ad:
Thanks to @jorgepiloto for nudging me on to investigate this further.
I lucky ran into a github issue on some other repo, we led me to another similar issues in some other repo, which led me to ... you get the point 😆. People have this issue in the past, with different symptoms.
- https://github.com/blacklanternsecurity/bbot/issues/1563
- https://github.com/nedbat/coveragepy/issues/1310
- https://github.com/pytest-dev/pytest/issues/13137
- https://github.com/pytest-dev/pytest-cov/issues/261
- https://github.com/pytest-dev/pytest/issues/9652
- https://github.com/blacklanternsecurity/bbot/issues/1571

My assessment of all information I could digest from those issues => the easiest thing to try is deactivate the pytest plugins, as those are the likely culprits causing hanging.

Deactivating pytest-cov and not dumping the junitxml files led to these results:
image

Looks like we have identified the culprit for starters.

moe-ad:
Running without pytest-cov only:

image

So coverage seems like the culprit.

moe-ad:
I can see we are still on pytest-cov 5.0.0, let's bump that to 6.0.0 to see if that fixes things.

moe-ad:
Yeah. pytest-cov is the culprit and not any of the other plugins. Another related issue:
- https://github.com/pytest-dev/pytest-cov/issues/159

moe-ad:
Adding this issue too:
- https://github.com/pytest-dev/pytest-cov/issues/669

PProfizi:
> Yeah. pytest-cov is the culprit and not any of the other plugins. Another related issue:
> 
> * [4-5x slowdown using pytest-cov pytest-dev/pytest-cov#159](https://github.com/pytest-dev/pytest-cov/issues/159)

Thanks @moe-ad, does that mean we cannot have coverage with the proposed changes?

moe-ad:
@jorgepiloto @PProfizi looks like we are good to go.

tox.ini
```diff
@@ -2,83 +2,128 @@
...
 
 depends = pretest
 
 setenv =
     # Pytest extra arguments
-    COVERAGE = --cov=ansys.dpf.core --cov-report=xml --cov-report=html --log-level=ERROR --cov-append
+    COVERAGE = --cov=ansys.dpf.core --log-level=ERROR
     RERUNS = --reruns=2 --reruns-delay=1
     DEBUG = -v -s --durations=10 --durations-min=1.0
+    COVERAGE_FILE = {work_dir}/.cov/.coverage.{env_name}
```
jorgepiloto, line 126: Excellent! So using `{env_name}` allows us to create a cov file with the name of the environment that created it, right? This is great! It avoids passing environment variables.

tox.ini
```diff
@@ -101,13 +148,26 @@ setenv =
...
-deps = 
-    -r requirements/requirements_test.txt
+    TEMP = {env_tmp_dir}
+    TMP = {env_tmp_dir}
 
 commands =
-    pytest {env:PYTEST_PYTHON_FILES} {env:DEBUG} {env:COVERAGE} {env:RERUNS} {env:JUNITXML}
+    python -m pytest {env:PYTEST_PYTHON_FILES} {env:DEBUG} {env:RERUNS} {env:JUNITXML} {env:COVERAGE} {posargs}
+
+[testenv:report]
```
jorgepiloto, line 161: Feels too generic. What about?

```suggestion
[testenv:covreport]
```

jorgepiloto (changes requested):

jorgepiloto:
Thanks for the very big effort here, @moe-ad. All the research you performed is super valuable and will prevent us from this tricky issue in the future.

.github/workflows/tests.yml
```diff
@@ -191,18 +198,15 @@ jobs:
...
+          python -c "from ansys.dpf import core"
 
       - name: "Prepare Testing Environment"
         uses: ansys/pydpf-actions/prepare_tests@v2.3
         with:
           DEBUG: true
 
-      - name: "List installed packages"
-        shell: bash
-        run: pip list
```
PProfizi, line 204: @moe-ad do we still get somewhere the list of installed packages in the environments generated by tox? For easier debugging.

.github/workflows/tests.yml
```diff
@@ -211,154 +215,129 @@ jobs:
...
+            fi
+          fi
 
-      - name: "Set pytest arguments"
+      - name: "Organize test files"
         shell: bash
         run: |
-          echo "COVERAGE=--cov=ansys.dpf.${{env.MODULE}} --cov-report=xml --cov-report=html --log-level=ERROR --cov-append" >> $GITHUB_ENV
-          echo "RERUNS=--reruns 2 --reruns-delay 1" >> $GITHUB_ENV
+          tox -e pretest
```
PProfizi, line 239: @moe-ad so when run locally, do we need to manually run each of the following steps separately? Or can we use a single command?

.github/workflows/tests.yml
moe-ad: Tox will run pip freeze automatically on each of the environments e.g.,:
image

tox.ini
```diff
@@ -2,83 +2,128 @@
 
 # Usage instructions:
 
-# `tox` will run all tests sequentially, `tox --parallel` will run all tests in parallel (much faster).
-# Run specific selection of tests with `tox -e pretest,<list-of-tests>,posttest` e.g., `tox -e pretest,test-api,test-launcher,posttest`
-# `--parallel` flag can be passed when running specific selections.
+# `tox` will run all "envlist" tests sequentially, `tox --parallel` same tests in parallel (much faster).
```
PProfizi, line 5: @moe-ad once this is ready for use we should also document this in the `Contributing` section of the documentation.

tox.ini
```diff
@@ -2,83 +2,128 @@
...
 
-    # Organize test files
+[testenv:pretest]
+description = Environment to organize test files prior to testing
+
+skip_install = True
+
+deps =
+
+commands_pre =
```
PProfizi, line 89: @moe-ad why do we have to declare them even though they are empty?

moe-ad: This is to make overriding testing config easier from the command line.
The idea is that base `[testenv]` config is inherited by child environments (`[testenv:A]`, `[testenv:B]` etc.), unless overridden in the child environments.

When TOX_EXTRA_ARG is set in tests.yml, it targets some `[testenv]` config => child environments will inherit this config. So in environments where we do not want this to be inherited, we specify an empty value.

jorgepiloto (approved):
This is impressive. Just approved it. Thanks a lot, @moe-ad.

PR #2040, opened 2025-01-27
Fix Scoping.ids setter for int64 arrays

### LINKED ISSUE #2037: DPF scoping creation using Numpy arrays in Numpy2

### Description of the bug

When creating a scoping from Numpy arrays in DPF, there is a significant difference in behavior in Numpy 1 vs. Numpy 2. A minimum example is as follows:
```
import numpy as np

nids = np.array([1,2,3,4],dtype=int)
scop = dpf.Scoping(ids=nids, location= dpf.locations.nodal)
print(nids)
print(scop.ids)
```

Using Numpy 1 (1.26.4):

Image

Using Numpy 2 (2.2.2, the default version when installing DPF 0.13.4 today):

Image

One can see that the scoping generated from Numpy2 array is incorrect (it adds 0 padding every other entry), while this is correct in Numpy 1. This is a dangerous behavior since if a new user installs DPF in a fresh virtual environment, it automatically pulls the latest version of Numpy.

### Steps To Reproduce

1. Install DPF and Numpy2 in virtual environment
2. Run the following example code

```
import numpy as np

nids = np.array([1,2,3,4],dtype=int)
scop = dpf.Scoping(ids=nids, location= dpf.locations.nodal)
print(nids)
print(scop.ids)
```

Windows

Ansys 2025 R1

3.10, 3.13

### Installed packages

The working version:
ansys-dpf-core==0.13.4
cachetools==5.5.1
certifi==2024.12.14
charset-normalizer==3.4.1
colorama==0.4.6
google-api-core==2.24.0
google-api-python-client==2.159.0
google-auth==2.38.0
google-auth-httplib2==0.2.0
googleapis-common-protos==1.66.0
grpcio==1.70.0
httplib2==0.22.0
idna==3.10
importlib_metadata==8.6.1
numpy==1.26.4
packaging==24.2
proto-plus==1.25.0
protobuf==5.29.3
psutil==6.1.1
pyasn1==0.6.1
pyasn1_modules==0.4.1
pyparsing==3.2.1
requests==2.32.3
rsa==4.9
tqdm==4.67.1
uritemplate==4.1.1
urllib3==2.3.0
zipp==3.21.0

The default:
ansys-dpf-core==0.13.4
cachetools==5.5.1
certifi==2024.12.14
charset-normalizer==3.4.1
colorama==0.4.6
google-api-core==2.24.0
google-api-python-client==2.159.0
google-auth==2.38.0
google-auth-httplib2==0.2.0
googleapis-common-protos==1.66.0
grpcio==1.70.0
httplib2==0.22.0
idna==3.10
importlib_metadata==8.6.1
numpy==2.2.2
packaging==24.2
proto-plus==1.25.0
protobuf==5.29.3
psutil==6.1.1
pyasn1==0.6.1
pyasn1_modules==0.4.1
pyparsing==3.2.1
requests==2.32.3
rsa==4.9
tqdm==4.67.1
uritemplate==4.1.1
urllib3==2.3.0
zipp==3.21.0

PProfizi:
Closes #2037 

Edit: this fixes the issue for `np.ndarray(dtype=int64)` which is the new default, however for some reason the case:

```py
import numpy as np

nids = np.array([1,2,3,4],dtype=int)
nids = list(nids)
print(nids)
scop = dpf.Scoping(ids=nids, location= dpf.locations.nodal)
print(nids)
print(scop.ids)
```

already worked despite the list being:
`[np.int64(1), np.int64(2), np.int64(3), np.int64(4)]`

Why would this case work? In theory these too would need to be converted to `int32`.


src/ansys/dpf/core/scoping.py
```diff
@@ -191,6 +194,9 @@ def _set_ids(self, ids: IdVectorType):
         """
         if isinstance(ids, range):
             ids = list(ids)
+        if isinstance(ids, np.ndarray):
+            if ids.dtype == np.int64:
```
cbellot000, line 198: shouldn't we rather check the opposite: if not int32 -> convert to int32?

rafacanton: Yes, can we have any other type of int in numpy? I mean, for sure the case of an array of floats shouldn't be handled, but any int should be handled as int32

PProfizi, line 198:
```suggestion
            if ids.dtype != np.int32:
```

rafacanton: @PProfizi and what if I pass the array `np.array([0.2, 0.6, 5.0])`? That should raise, I think.

rafacanton (approved):

src/ansys/dpf/core/scoping.py
```diff
@@ -49,11 +49,12 @@
 
 if TYPE_CHECKING:  # pragma: nocover
     from ctypes import c_void_p as ScopingPointer
+    from numpy import typing as np_typing
```
PProfizi, line 52:
```suggestion

    from numpy import typing as np_typing
```

PR #2045, opened 2025-01-29
Add result_name pin in custom and modify example

oparreno:
- Add `result_name `pin in `custom `operator.
- Modify `02-hdf5_serialize_and_read` example to use the source operator instead.


rafacanton:
@oparreno The operators code is autogenerated, we shouldn't modify it manually. Changes will be ported from the server when the hdf5 issue is fixed

rafacanton (approved):

PR #2059, opened 2025-02-07
Pin aliases

PProfizi:
This PR exposes APIs for pin name aliases in PyDPF-Core, along with a template update for operator module generation to enable retro-compatibility of scripts using aliases, as well as warning about deprecation of said aliases.

### DIFF
--- a/src/ansys/dpf/core/operator_specification.py
+++ b/src/ansys/dpf/core/operator_specification.py
@@ -78,6 +78,7 @@ class PinSpecification:
     optional: bool
     ellipsis: bool
     name_derived_class = str
+    aliases: list[str]
 
     def __init__(
         self,
@@ -87,6 +88,7 @@ class PinSpecification:
         optional=False,
         ellipsis=False,
         name_derived_class="",
+        aliases=[],
     ):
         self.name = name
         self.type_names = type_names
@@ -94,6 +96,7 @@ class PinSpecification:
         self.document = document
         self.ellipsis = ellipsis
         self.name_derived_class = name_derived_class
+        self.aliases = aliases
 
     @property
     def type_names(self) -> list[str]:
@@ -140,6 +143,7 @@ class PinSpecification:
             other.optional,
             other.ellipsis,
             other.name_derived_class,
+            other.aliases,
         )
 
     def __repr__(self):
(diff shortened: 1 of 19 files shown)


PProfizi:
Here is an example of a server sync update with the current changes (plus a few others due to diff with master):
https://github.com/ansys/pydpf-core/pull/2067/files

rafacanton:
@PProfizi This looks nice, also the operator code update. Can we add a test to ensure that we get the warning when we try to connect `op.inputs.ponderation`? Of course, when we eliminate ponderation in the future, the test will crash and we just remove it.

src/ansys/dpf/core/operator_specification.py
```diff
@@ -429,7 +433,16 @@ def _fill_pins(self, binput, to_fill):
                     self._api.operator_specification_get_pin_type_name(self, binput, i_pin, i_type)
                     for i_type in range(n_types)
                 ]
-
+                pin_aliases = []
+                if server_meet_version("10.0", self._server):
```
PProfizi, line 437:
```suggestion
                if hasattr(self._api, "operator_specification_get_pin_num_aliases"):
```

src/ansys/dpf/core/operator_specification.py
```diff
@@ -429,7 +433,16 @@ def _fill_pins(self, binput, to_fill):
...
+                    for i_alias in range(
+                        self._api.operator_specification_get_pin_num_aliases(self, binput, i_pin)
+                    ):
+                        pin_aliases.append(
+                            self._api.operator_specification_get_pin_alias(
+                                self, binput, i_pin, i_alias
+                            )
+                        )
                 pin_derived_class_type_name = ""
                 if server_meet_version("7.0", self._server) and hasattr(
```
rafacanton, line 447: When @MichaelNale added this we were checking that the method is in the api and that the server version was met. Now that I see your solution, the name check seems enough. Can you remove the server version check here? Thanks!

PProfizi:
> @PProfizi This looks nice, also the operator code update. Can we add a test to ensure that we get the warning when we try to connect `op.inputs.ponderation`? Of course, when we eliminate ponderation in the future, the test will crash and we just remove it.

I'd actually like to merge this without having to wait for the PR server-side which does the switch of `ponderation` to an alias.
I will add the test in a future PR. See #2117

rafacanton (approved):

src/ansys/dpf/core/operator_specification.py
PProfizi: It turns out this is necessary

rafacanton: And why is it not needed for the alisases?

rafacanton (approved):

PR #2072, opened 2025-02-11
fix: requirements

jorgepiloto:
Removes empty file and `black` as a dependency in favor of Ruff.

### DIFF
--- a/requirements/requirements_build.txt
+++ b/requirements/requirements_build.txt
@@ -1,4 +1,3 @@
-black==25.1.0
 build==1.2.2.post1
 chevron==0.14.0
 wheel==0.45.1


requirements/requirements_build.txt
```diff
@@ -1,4 +1,3 @@
-black==25.1.0
```
PProfizi, line 1: Hi @jorgepiloto, again, this is used in the `ci/code_generation.py` script. We cannot remove it here or we need to also modify the file mentioned.

PProfizi (changes requested):
Cannot remove `black` directly.

moe-ad (dismissed):
LGTM!

requirements/requirements_build.txt
jorgepiloto: Hi @PProfizi, I do not see the tool being used in the script you refer to https://github.com/ansys/pydpf-core/blob/master/.ci/code_generation.py

PProfizi: @jorgepiloto sorry I meant [here](https://github.com/ansys/pydpf-core/blob/a5384553348b86393a6c169a7872ceb8a407e8e3/src/ansys/dpf/core/operators/build.py#L168), which is what runs when we call `code_generation.py`. @moe-ad FYI

PProfizi: We basically use it to reformat some code we autogenerate based on a template.

moe-ad:
@jorgepiloto see #2064.

requirements/requirements_build.txt
jorgepiloto: Thanks for explanation, @PProfizi. @moe-ad, what do you think it is possible to explore Ruff's API for this? I am not sure if this tool has an API for Python. If it had, we could completely drop black.

Issue #2090, opened 2025-02-17
Unstable 232 retro tests


PProfizi:
### Before submitting the issue

- [x] I have checked for [Compatibility](https://dpf.docs.pyansys.com/version/stable/getting_started/compatibility.html#id1) issues
- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

Several PRs have seen failures on 232 ubuntu due to bad port in use detection.
It is a known issue that is some specific contexts (for example on Ubuntu for 232), the automatic free port detection fails due to ports used by a server shutting down being reported as free while they are not yet free.
It is not usually an issue, but during testing as in the pipelines, this creates random failures such as here:
https://github.com/ansys/pydpf-core/pull/1944 

We need to improve stability of the test job by adding fixtures in dedicated test modules to improve server port management. Basically handling port assignment manually rather than automatically.

### Steps To Reproduce

```
E0108 13:38:01.093852653    7465 server_chttp2.cc:40]        {"created":"@1736343481.093828528","description":"No address added out of total 1 resolved","file":"/home/staff/kkripa/tfsagent/_work/20/b/.conan/data/grpc_base/1.25.0/thirdparty/stable/build/1036933dfdff90461d4bf4154f59e6aa78392d87/grpc/src/core/ext/transport/chttp2/server/chttp2_server.cc","file_line":394,"referenced_errors":[{"created":"@1736343481.093822727","description":"Unable to configure socket","fd":25,"file":"/home/staff/kkripa/tfsagent/_work/20/b/.conan/data/grpc_base/1.25.0/thirdparty/stable/build/1036933dfdff90461d4bf4154f59e6aa78392d87/grpc/src/core/lib/iomgr/tcp_server_utils_posix_common.cc","file_line":208,"referenced_errors":[{"created":"@1736343481.093818479","description":"Address already in use","errno":98,"file":"/home/staff/kkripa/tfsagent/_work/20/b/.conan/data/grpc_base/1.25.0/thirdparty/stable/build/1036933dfdff90461d4bf4154f59e6aa78392d87/grpc/src/core/lib/iomgr/tcp_server_utils_posix_common.cc","file_line":181,"os_error":"Address already in use","syscall":"bind"}]}]}
Fatal Python error: Segmentation fault
```

### Which Operating System causes the issue?

Linux

### Which DPF/Ansys version are you using?

Ansys 2023 R2

### Which Python version causes the issue?

3.9

### Installed packages

NA

[...]

PR #2095, opened 2025-02-18
Explicit server port allocation in tests

PProfizi:
Add get_next_port() function to conftest for usage in fixtures providing gRPC servers.

trying to solve #2090

Edit: looks like the error still happens when running [examples on Docker](https://github.com/ansys/pydpf-core/actions/runs/13393312510/job/37406054387?pr=2095#step:14:180)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -438,6 +438,15 @@ configsserver_type, config_namesserver_type = remove_none_available_config(
 )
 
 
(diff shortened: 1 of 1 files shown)


moe-ad (approved):
@PProfizi looks like the docker failure was just an instability. 
https://github.com/ansys/pydpf-core/actions/runs/13407780206/job/37450854447?pr=2095

Issue #928, opened 2023-05-03
OSError on wrong output label when connecting two workflows


ClappeB:
### Before submitting the issue

- [X] I have checked for [Compatibility](https://dpf.docs.pyansys.com/getting_started/compatibility.html#id1) issues
- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

When connecting two workflows together, if specifying wrong output name for distant workflow to connect on local workflow input, the code will throw an "OSError: exception: access violation reading 0x0000000000000000".
For example, let be:
- Workflow1 with output "mesh_output"
- Workflow2 with input "mesh_input"

If doing `Workflow2.connect_with(Workflow1, {"mesh_out": "mesh_input"})` and running the workflow, this will throw OSError (because "mesh_out" doesn't exist on Workflow1).

Value fix for below: USING PYTHON 3.11 (not available when creating the issue)

### Steps To Reproduce

```python
from ansys.dpf import core as dpf
from ansys.dpf.core import examples

workflow1 = dpf.Workflow()
mesh_provider = dpf.operators.mesh.mesh_provider(data_sources=dpf.DataSources(result_path=examples.find_multishells_rst()))
workflow1.add_operator(mesh_provider)
workflow1.set_output_name("mesh_output", mesh_provider.outputs.mesh)

workflow2 = dpf.Workflow()
skin_op = dpf.operators.mesh.skin()
workflow1.add_operator(skin_op)
workflow2.set_input_name("mesh_input", skin_op.inputs.mesh)
workflow2.set_output_name("skin_mesh", skin_op.outputs.mesh)

workflow2.connect_with(workflow1, {"mesh_out": "mesh_input"})   \# Wrong output name
workflow2.get_output("skin_mesh", dpf.types.meshed_region)
```

### Which Operating System are you using?

Windows

### Which DPF/Ansys version are you using?

Ansys 2023 R1

### Which Python version are you using?

3.10

### Installed packages

ansys-api-platform-==1.0.0b3
ansys-dpf-core==0.8.2.dev0
ansys-dpf-flowdiagram==0.1.0.dev0 D:\repos\flow-diagram\backend
ansys-dpf-gate==0.3.1
ansys-dpf-gatebin==0.3.1
ansys-grpc-dpf==0.7.1
ansys-platform-instancemanagement==1.1.1
anyio==3.6.2
cachetools==5.3.0
certifi==2022.12.7
charset-normalizer==3.1.0
click==8.1.3
[...]

PR #2110, opened 2025-02-21
Raise useful error on workflow.connect_with for wrong labels

PProfizi:
Solves #928 

PR on PyDPF-Post will make use of new `Workflow.safe_connect_with()`.

Edit: not sure about the naming. Should it rather be `Workflow.connect_with_available`, or something else? @BClappe 

Also I realize that changing the behavior of `Workflow.connect_with` to now raise if an input or output is unavailable, is a breaking change of a sort. Maybe we should actually keep `connect_with` as-is or even put the filtering logic there, and create a new method `Workflow.connect_with_error` which raises in the case of an unavailable input or output. What do you think?

### DIFF
--- a/src/ansys/dpf/core/workflow.py
+++ b/src/ansys/dpf/core/workflow.py
@@ -22,6 +22,8 @@
 
 """Workflow."""
 
+from __future__ import annotations
+
 from enum import Enum
 import logging
 import os
@@ -714,7 +716,11 @@ class Workflow:
         return out
 
     @version_requires("3.0")
-    def connect_with(self, left_workflow, output_input_names=None):
+    def connect_with(
+        self,
+        left_workflow: Workflow,
+        output_input_names: Union[tuple[str, str], dict[str, str]] = None,
+    ):
         """Prepend a given workflow to the current workflow.
 
         Updates the current workflow to include all the operators of the workflow given as argument.
@@ -724,9 +730,9 @@ class Workflow:
 
         Parameters
         ----------
-        left_workflow : core.Workflow
+        left_workflow:
             The given workflow's outputs are chained with the current workflow's inputs.
-        output_input_names : str tuple, str dict optional
+        output_input_names:
             Map used to connect the outputs of the given workflow to the inputs of the current
             workflow.
             Check the names of available inputs and outputs for each workflow using
--- a/src/ansys/dpf/core/workflow.py
+++ b/src/ansys/dpf/core/workflow.py
@@ -715,40 +715,12 @@ class Workflow:
             out.append(self._api.work_flow_output_by_index(self, i))
         return out
 
-    def safe_connect_with(
-        self,
-        left_workflow: Workflow,
-        output_input_names: Union[tuple[str, str], dict[str, str]],
-    ):
-        """Prepend a given workflow to the current workflow for valid connections only.
-
-        See Workflow.connect_with for more information on the connection logic.
-
-        Parameters
-        ----------
-        left_workflow:
-            The given workflow's outputs are chained with the current workflow's inputs.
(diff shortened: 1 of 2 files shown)


src/ansys/dpf/core/workflow.py
```diff
@@ -713,8 +715,41 @@ def output_names(self):
...
+            output_input_names = {output_input_names[0]: output_input_names[1]}
+        valid_connections = dict(
+            filter(
+                lambda item: item[0] in left_workflow.output_names and item[1] in self.input_names,
+                output_input_names.items(),
+            )
+        )
+        self.connect_with(left_workflow, valid_connections)
+
     @version_requires("3.0")
```
rafacanton, line 747: The decorator for version should also be passed to the new method, right?

rafacanton:
@PProfizi To answer the question in the description, the original problem is that before we were crashing in the kernel and not raisin in Python, so in any way the connection with wrong labels was not warking (the process couldn't be run end-to-end). Therefore, why having a new method?

PProfizi:
> @PProfizi To answer the question in the description, the original problem is that before we were crashing in the kernel and not raisin in Python, so in any way the connection with wrong labels was not warking (the process couldn't be run end-to-end). Therefore, why having a new method?

Because weirdly it did not crash all the time, for example in PyDPF-Post's tests I have to update calls because workflow connections with unavailable inputs or outputs are done all the time.
I think the case where it crashes is a specific case.

PProfizi:
Just to sum-up:

- initial reported problem is that `Workflow.connect_with()` sometimes crashes the server due to unavailable inputs/outputs on either workflow being connected -> instead we should check for that in Python before the server call, and raise an explicit and helpful Python ValueError.
- This however started raising errors in PyDPF-Post where calls to `Workflow.connect_with()` which used to work now raise the error, meaning they did not crash the server despite unavailable inputs/outputs -> this means the change is not only a bug fix but also a breaking change for cases where it used to work and may have to be treated as such
- A solution is to filter the inputs/outputs to connect given to the method, and only perform connections for available ones. This way instead of raising an error we filter and clean the dict for the user -> this is however hidden from the user who may not understand why a connection is not made. Should we then raise a warning? These are going to be raised a lot.
- This is why another solution is to have two distinct methods: one which filters for the user, and one which raises if something is unavailable. -> we could also add a new option to `Workflow.connect_with()` which switches between filtering and raising.

rafacanton:
> Just to sum-up:
> 
> * initial reported problem is that `Workflow.connect_with()` sometimes crashes the server due to unavailable inputs/outputs on either workflow being connected -> instead we should check for that in Python before the server call, and raise an explicit and helpful Python ValueError.
> * This however started raising errors in PyDPF-Post where calls to `Workflow.connect_with()` which used to work now raise the error, meaning they did not crash the server despite unavailable inputs/outputs -> this means the change is not only a bug fix but also a breaking change for cases where it used to work and may have to be treated as such
> * A solution is to filter the inputs/outputs to connect given to the method, and only perform connections for available ones. This way instead of raising an error we filter and clean the dict for the user -> this is however hidden from the user who may not understand why a connection is not made. Should we then raise a warning? These are going to be raised a lot.
> * This is why another solution is to have two distinct methods: one which filters for the user, and one which raises if something is unavailable. -> we could also add a new option to `Workflow.connect_with()` which switches between filtering and raising.

Thanks for summarizing everything so clearly. I think that the followed approach (with the 2 methods) is the right one giving all the compatibility restrictions

PProfizi:
@rafacanton I ended-up adding the `permissive` parameter with a default value of `True` which I think is the graceful way of solving this.

rafacanton (approved):

PR #2115, opened 2025-02-25
Feat: support Markdown and LaTeX in operator specs

PProfizi:


.ci/code_generation.py
```diff
@@ -8,6 +8,9 @@
 from ansys.dpf import core
 from ansys.dpf.core.operators import build
 
+core.set_default_server_context(core.AvailableServerContexts.premium)
+core.start_local_server(config=core.AvailableServerConfigs.LegacyGrpcServer)
```
PProfizi, line 12: This prevents deletion of operators if for whatever reason we failed to start a DPF server.

PProfizi:
@ansys-akarcher I checked the generated documentation and there is no change for current text-only descriptions.
The only impact this PR currently has is to make the operator code generation a bit longer (5min instead of 1min approximately).

src/ansys/dpf/core/operators/build.py
```diff
@@ -11,18 +12,15 @@
...
     docstring = ""
-    if specification.description:
-        docstring += "\n".join(
-            wrap(specification.description, subsequent_indent="    ")
-        )
-        docstring += "\n\n"
-    docstring = docstring.rstrip()
-    return docstring.replace('"', "'")
+    if specification_description:
+        docstring += specification_description.replace("\n", "\n    ")
```
ansys-akarcher, line 22: Not sure to understand the whitespaces offsets l.22 and l.85

ansys-akarcher (approved):

src/ansys/dpf/core/operators/build.py
PProfizi: @ansys-akarcher This is to follow the right indentation based on where we insert it in the template.

PR #2141, opened 2025-03-10
DpfVector BUG: DpfVector were created on one clayer and deleted in another 

cbellot000:
When working on distributed workflows (in process and gRPC), DPFVector were sometimes created on one clayer and deleted on another. This prevented any changes in the DPFVector.


PProfizi:
@cbellot000 could we maybe have a test to ensure behavior is fixed?

cbellot000:
> @cbellot000 could we maybe have a test to ensure behavior is fixed?

We do have a test: it's the reason why https://github.com/ansys/pydpf-core/pull/2135 is failing :)

PProfizi (approved):

PR #2160, opened 2025-03-14
ci: drop py39 in pipelines and add 3.12

moe-ad:


moe-ad:
Triggered release ci workflow: https://github.com/ansys/pydpf-core/actions/runs/13858708918

PR #2161, opened 2025-03-14
Cyclic examples modifications to set up cyclic operators private

oparreno:
Modify old calls to cyclic_expanded_result to result in examples. 
Associated pipelines with TFS code: 
https://github.com/ansys/pydpf-core/actions/runs/13856464751/job/38778362266?pr=2159

### DIFF
--- a/examples/04-advanced/00-multistage_advanced_options.py
+++ b/examples/04-advanced/00-multistage_advanced_options.py
@@ -64,10 +64,11 @@ print(
 
 
 # Create displacement cyclic operator
-UCyc = dpf.operators.result.cyclic_expanded_displacement()
+UCyc = dpf.operators.result.displacement()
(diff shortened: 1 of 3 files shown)


rafacanton (approved):

PProfizi:
Hi @oparreno I have the release CI failing on retro with 251 [here](https://github.com/ansys/pydpf-core/actions/runs/13883666945/job/38845483687#step:14:116)
I think `ansys\dpf\core\operators\metadata\cyclic_support_provider` has changed internal name, right?

oparreno:
@PProfizi, yes for `cyclic_support_provider` the name was changed from `mapdl::rst::cyclic_support_provider`. In this case, we can't really re-record it with the previous name as the original name exists as an mapdl operator.

PR #2167, opened 2025-03-18
Ci: add test analytics

PProfizi:
This PR adds test result upload to Codecov to start getting test analytics (test failure rates and average duration over commits for dev branches and over time for the main branch).
This will ease identification of flaky tests and prioritization of investigations for random failures.


PProfizi:
@moe-ad it looks like it is working 
image

PR #2203, opened 2025-04-03
Add datasources namespace API

cbellot000:


rafacanton (approved):

tests/test_datasources.py
```diff
@@ -25,6 +25,7 @@
 import pytest
 
 from ansys import dpf
+from ansys.dpf.core import examples
 import conftest
 
 skip_always = pytest.mark.skipif(True, reason="Investigate why this is failing")
```
PProfizi, line 31: @rafacanton @cbellot000 It looks like this test module is skipped for some reason

Issue #2201, opened 2025-04-01
Performance of get/set field data


FedericoNegri:
### Before submitting the issue

- [x] I have checked for [Compatibility](https://dpf.docs.pyansys.com/version/stable/getting_started/compatibility.html#id1) issues
- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

Some recent changes (https://github.com/ansys/pydpf-core/commit/f7be394515266766b6208ef4f341a5fdfec558db?) seem to have slowed down manipulating field data (by orders of magnitude!).

### Steps To Reproduce

This is a minimal script to reproduce the issue. Using `uv` I could easily run it with different "versions" of PyDPF-Core and compare timings.

Using the latest released version (0.13.6) it takes ~20ms to run. With current main it instead takes >5s.

Note: I'm aware the operations tested here could be done in a different way, by setting the field data and scoping outside of the loop. But still, I feel like this could indicate some bigger problem.

```python
# /// script
# requires-python = "==3.12.9"
# dependencies = [
#     "numpy==2.2.4",
#     "ansys-dpf-core==0.13.6", #fast
#     #"ansys-dpf-core @ git+https://git@github.com/ansys/pydpf-core.git@e2a1cf0caa01d3376e956cbea026590a40ccbf48", #fast
#     #"ansys-dpf-core @ git+https://git@github.com/ansys/pydpf-core.git@f7be394515266766b6208ef4f341a5fdfec558db", #slow
# ]
# ///
"""
Script to measure performance of creating a DPF field.

How to:

1. Install uv if not already installed:
    pip install uv
   See https://docs.astral.sh/uv/getting-started/installation/#standalone-installer

2. Save this code as, say, `create_field.py` and run it:
    uv run create_field.py

3. Change the ansys-dpf-core version in the script header and re-run the script:
    uv run create_field.py
"""

import time
import numpy as np
from ansys.dpf.core import fields_factory

from ansys.dpf import core as dpf

server = dpf.start_local_server(
    context=dpf.AvailableServerContexts.premium,
    config=dpf.AvailableServerConfigs.InProcessServer,
    as_global=True,
)

num_entities = int(1e+6)
field = fields_factory.create_scalar_field(
    num_entities=num_entities, location=dpf.locations.elemental, server=server
)
field.name = "my_field"
field.data = np.zeros(num_entities, dtype=np.int32)
field.scoping.ids = np.zeros(num_entities, dtype=np.int32)

[...]

PR #2249, opened 2025-04-30
Fix performance of DPF vector

cbellot000:
closing #2201 

DPF vector became slow because they were always checking if the modification of the vector by
1. making a deep copy of the vector
2. comparing the whole vector

**Why performance changed between 0.13.6 and master?**
Before 0.13.7.dev, it was hard coded that in process dpf vector wouldn't "check modification" and though bypassed dpf server logic. 0.13.7.dev is now relying on DPF server logic for updating or not dpf vector.

**Proposed fixes**
- DPFVector are now always "committed" to dpf server whether they have been modified or not (as checking for modification is very slow)
- Performance is highly improved for server version checks by caching previous checks

**Benchmark**
Test is taken for #2201, changing params are:
- pydpf-core version
- server types: in process/gRPC
- vector operations: data is only "gotten", or data is "gotten" and "set"

| Server type | Vec operation |      0.13.6    |  0.13.7.dev |
|----------|:----------|:-------------:|------:|
| In Process | Get |  0.009 s | 0.008 s |
| In Process | Get/Set |  0.012 s |   0.011 s |
| gRPC | Get  |  6.608 s |   3.773 s |
| gRPC | Get/Set| 8.707 s |    3.707 s |


src/ansys/dpf/gate/dpf_vector.py
```diff
@@ -33,13 +33,18 @@ class DPFVectorBase:
 
     def __init__(self, owner, api):
         self.dpf_vector_api = api
-        self._modified = False
-        self._check_changes = True
+
+        # The updated version of the DPF vector will always be committed to DPF.
+        # Ideally, this should be set to True only when modified, however this is not possible to do that efficiently.
+        # Consequently, for performance reasons, it's much better to always commit the vector to DPF rather than
```
PProfizi, line 39: @cbellot000 is this also true in the case of a huge vector in gRPC on a remote machine with a high ping?

cbellot000: since it's twice faster, I think that it should be safe to assume that it will as fast

rafacanton (approved):

PR #2252, opened 2025-04-30
Comment failing tests for geometry helper

PProfizi:
Removing tests in test_geometry.py on create_plane_from_lines as this helper is wrong anyway and this is blocking pyvista updates where a check has been added.
Hesitating with pure deprecation of the helper as defining a plane from two lines (four points) is undefined behavior anyway.

### DIFF
--- a/tests/test_geometry.py
+++ b/tests/test_geometry.py
@@ -207,31 +207,31 @@ def test_create_plane_from_points(points):
     plane.plot()
 
 
-plane_lines_data = [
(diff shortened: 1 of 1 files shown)


rafacanton:
@PProfizi 2 lines define a plane in 3D space only:
- If they intersect.
- If they are parallel.
Otherwise (if they cross in 3D space), no plane can be defined... The API still makes sense, but it should give a meaningful error if not in the cases above

PR #2272, opened 2025-05-12
ci: make pydpf-post job information-only

PProfizi:
The PyDPF-Core CI runs PyDPF-Post tests for information purposes thanks to its `pydpf-post.yml` workflow.
This PR makes sure that the workflow is always reported as successful to not stop people from merging, while still [reporting errors](https://github.com/marketplace/actions/continue-on-error-comment) for PyDPF-Post tests as comments in the PR if any.
This is most useful for `server-sync` PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,4 +253,5 @@ jobs:
       post_branch: "master"
       standalone_suffix: ${{needs.pick_server_suffix.outputs.suffix}}
       test_docstrings: "true"
+      continue-on-error: true
     secrets: inherit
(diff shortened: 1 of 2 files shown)


PProfizi:
Here is a CI run on the daily standalone (where pydpf-post tests fail), still reporting PyDPF-Post workflow as passing https://github.com/ansys/pydpf-core/actions/runs/14966392212

PProfizi:
@rafacanton the `continue-on-error` already makes the workflow go green despite PyDPF-Post tests failing as you can see in the CI in the previous comment.
I just added use of a `comment` action, which we can test by merging this and then updating the daily `server-sync` PR currently failing, to see if comments do appear.
In any case it will have solved the initial issue.

rafacanton (approved):

PR #2278, opened 2025-05-12
Add missing homogeneities

rafacanton:
Add missing homogeneities defined in the server but not in the client. Correct the value of the `current` homogeneity, which was mistaken by `current_density`.


PProfizi (approved):

PProfizi:
@rafacanton wouldn't it be better if this Enum was provided by the framework?

rafacanton:
> @rafacanton wouldn't it be better if this Enum was provided by the framework?

Yes, we discussed it about 1 month ago. We should fill this enum automatic upon code generation update, so it is entirely on the server. I created an operator to do that :)

Issue #2274, opened 2025-05-12
Invalid vector subscript after DPF Vector performance improvement


janvonrickenbach:
### Before submitting the issue

- [x] I have checked for [Compatibility](https://dpf.docs.pyansys.com/version/stable/getting_started/compatibility.html#id1) issues
- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

I get the following error if I try to evaluate strains from a composite model (see script below)

ansys.dpf.gate.errors.DPFServerException: a 'data processing core error' error occurred: EPEL:155<-mapdl::rst::EPEL:176<-invalid vector subscript, failed to read elastic strains from rst file. 

This has worked before https://github.com/ansys/pydpf-core/commit/677b743c984603e4392e82d59a8cb5d193757d19 was merged. 

Notes:

* I tried to make a more minimal model, but did not succeed. One observation: The script runs correctly if line 220 in ansys.dpf.composites.layup_info/_element_info.py is commented.

### Steps To Reproduce

Extract the attached folder and point to it in the result_folder variable. Install pydpf-composites (pip install ansys-dpf-composites). Start a dpf server and use the correct port in the script. The problem occurs only with a grpc server (remove the port argument below to use an in-process server).
```
from ansys.dpf import core as dpf
grpc_server_config = dpf.AvailableServerConfigs.GrpcServer
grpc_server = dpf.start_local_server(config=grpc_server_config)
grpc_server.info
```

then

```
from ansys.dpf.composites.composite_model import CompositeModel
from ansys.dpf.composites.server_helpers import connect_to_or_start_server
from ansys.dpf.composites.data_sources import (
    get_composite_files_from_workbench_result_folder,
)

server = connect_to_or_start_server(port=50058)


result_folder = r"composites"
files = get_composite_files_from_workbench_result_folder(result_folder)
composite_model = CompositeModel(files, server=server)

composite_model.core_model.results.elastic_strain().eval()
```

[composites.zip](https://github.com/user-attachments/files/20160321/composites.zip)

### Which Operating System causes the issue?

Windows

### Which DPF/Ansys version are you using?

DPF Server 2025.2.pre0

### Which Python version causes the issue?

3.10

### Installed packages

ansys-tools-path==0.7.1
anyio==4.9.0
asyncio-atexit==1.0.1
backoff==2.2.1
certifi==2025.4.26
cfgv==3.4.0
charset-normalizer==3.4.2
click==8.1.8
[...]

PR #2285, opened 2025-05-14
Fix update of empty vectors

cbellot000:
Empty (nullptr) vectors can be updated to non empty vectors of size 0 when using gRPC.
This can, for example, cause a property field to have a non empty data pointer after accessing the data pointer in python.

closing #2274


tests/test_dpf_vector.py
```diff
@@ -58,3 +58,45 @@ def test_perf_vec_getters(server_type):
...
+    field = dpf.Field(server=server_type)
+    field.data = np.zeros((100), dtype=np.double)
+    field.scoping.ids = list(range(1, 100))
+    assert np.allclose(field.get_entity_data(1), [0])
+    dp = field._data_pointer
+    dp = None
+    assert np.allclose(field.get_entity_data(1), [0])
+
+
+@conftest.raises_for_servers_version_under("5.0")
```
rafacanton, line 83: @cbellot000 6.0 to also cover 2023R1?

PR #2292, opened 2025-05-20
ci(workflow): optimize syntax

jorgepiloto:
This pull-request takes advantage of `matrix` jobs to short the workflow files, enhancing maintenance and readability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,7 @@ jobs:
       wheelhouse: false
       standalone_suffix: ${{needs.pick_server_suffix.outputs.suffix}}
       test_any: ${{ matrix.test-any }}
+      DOCSTRING: ${{ !matrix.test-any }}
     secrets: inherit
 
   docker_tests:


moe-ad:
Switching to ready for review to inspect action runs.

.github/workflows/ci.yml
```diff
@@ -179,68 +173,25 @@ jobs:
...
-  retro_222:
-    name: "retro 222"
+  tests-retro:
+    name: "Test DPF ${{ matrix.dpf.version }} compatibility"
     if: startsWith(github.head_ref, 'master') || github.event.action == 'ready_for_review' || !github.event.pull_request.draft
+    strategy:
+      fail-fast: false
+      matrix:
+        dpf:
+          - {"version": "251", "standalone-suffix": ""}
```
PProfizi, line 183: Hi @jorgepiloto is is possible to define all of this in repo variables?
I am starting to use them more and I was thinking about having an `ANSYS_VERSION_RETRO` variable such as `["252", "251", "242", "241"]`. I could add a `ANSYS_VERSION_RETRO_SUFFIX` variable. Would we be able to understand the list?

PProfizi: FYI https://github.com/ansys/pydpf-core/pull/2299

PProfizi: I could also refactor `tests.yml` and other workflow files so they understand `ANSYS_VERSION="241.sp01"`, and drop the use of the `standalone_suffix` variable.

moe-ad: > Hi @jorgepiloto is it possible to define all of this in repo variables? I am starting to use them more and I was thinking about having an `ANSYS_VERSION_RETRO` variable such as `["252", "251", "242", "241"]`. I could add a `ANSYS_VERSION_RETRO_SUFFIX` variable. Would we be able to understand the list?

I think a `fromJSON(var.ANSYS_VERSION_RETRO)` will do the trick for using such a list in a version definition. I am not certain, @jorgepiloto can also confirm.

moe-ad:
Triggered release workflow for testing validity:
https://github.com/ansys/pydpf-core/actions/runs/15343136

moe-ad (approved):

PR #2303, opened 2025-05-23
ci(issues): update issue templates for dev of DPF in 26R1

PProfizi:


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -87,13 +89,13 @@ body:
...
-       - '3.10'
-       - '3.11'
-       - '3.12'
        - '3.13'
+       - '3.12'
+       - '3.11'
+       - '3.10'
+       - '3.9'
+       - '3.8'
+       - '3.7'
```
moe-ad, line 98: We no longer support 3.9 and below right?

PProfizi: Hi @moe-ad yes I know but some users may be seeing issues on older versions of the client and/or server, which can be on Python 3.7.

PR #2322, opened 2025-06-02
Support PropertyField location after server changes

jvidalor:
Adter the last server changes in v11, the `PropertyField` location is defined on its `FieldDefinition` and it is independent from the location of its `Scoping`. Changes are made to the setters and getters to support this.

### DIFF
--- a/src/ansys/dpf/core/property_field.py
+++ b/src/ansys/dpf/core/property_field.py
@@ -91,7 +91,10 @@ class PropertyField(_FieldBase):
             field=property_field,
             server=server,
         )
-        self._field_definition_instance = self._load_field_definition()
+        if meets_version(server.version, "11.0"):
+            self._field_definition_instance = self._load_field_definition()
+        else:
+            self._field_definition_instance = None
 
     @property
     def _api(self) -> property_field_abstract_api.PropertyFieldAbstractAPI:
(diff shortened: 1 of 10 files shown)


src/ansys/dpf/core/property_field.py
```diff
@@ -91,7 +91,10 @@ def __init__(
             field=property_field,
             server=server,
         )
-        self._field_definition_instance = None
+        if meets_version(self._server.version, "11.0"):
+            self._field_definition_instance = self._load_field_definition()
```
cbellot000, line 95: why doing it at __init__ and not lazy anymore? It can have non negligeable impact on performance

src/ansys/dpf/gate/property_field_grpcapi.py
```diff
@@ -24,6 +24,16 @@ def csproperty_field_new_on_client(client, numEntities, data_size):
         request.datatype = "int"
         return field_grpcapi._get_stub(client).Create(request)
 
+    @staticmethod
+    def csproperty_field_new_location_on_client(client, numEntities, data_size, location):
```
cbellot000, line 28: why did we need a new API here, couldn't we do field_def_new and then set_location?

PProfizi (approved):

PR #2328, opened 2025-06-03
ci(tests): Add parallel_show_output=True to test environments in tox.ini

PProfizi:
This means that test steps will always show the list of tests which passed in parallel Tox environments, even if all are successful.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -54,6 +54,8 @@ pass_env =
 deps = 
     -r requirements/requirements_test.txt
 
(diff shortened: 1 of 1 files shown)


PProfizi:
Expected effect confirmed: https://github.com/ansys/pydpf-core/actions/runs/15418096278/job/43385649376?pr=2328#step:22:38

rafacanton (approved):

PR #2348, opened 2025-06-06
Test the FieldHeader getter operator

rafacanton:
A new operator has been added to be able to get the Header of a `Field` as a `DataTree`. This PR exposes and tests its getter


rafacanton:
This now fails as the server is not up-to-date. It passes with the right server: https://github.com/ansys/pydpf-core/pull/2344

src/ansys/dpf/core/field.py
```diff
@@ -804,6 +806,27 @@ def meshed_region(self) -> MeshedRegion:
...
+    @property
+    def header(self) -> DataTree:
+        """Field Header, which stores metadata of the Field.
+
+        Returns
+        -------
+        :class:`ansys.dpf.core.data_tree.DataTree`
+
+        """
+        server_meet_version_and_raise(
```
PProfizi, line 818: @rafacanton you can use `@version_requires` decorator instead, see [here for example](https://github.com/ansys/pydpf-core/blob/1095ea27673bcd5a51d1388d21bb3a88a071d6ba/src/ansys/dpf/core/collection_base.py#L125)

Sangsimon:
> @Sangsimon Putting this work on hold so it can be more generic. I think this should fall on the Header exposure you will be working on. TODOs:
> 
> * Expose a similar operator for StringFields (I don't know why I didn't do it, all pieces are there)
> * Expose the header getter here in the FieldBase class so all it children redirect the call to the appropriate op

We need to discuss this further so that I have more clarity on what we are trying to achieve with this generality.

Issue #2009, opened 2025-01-13
Including documentation from non-public modules.


moe-ad:
In versions of the api reference that used sphinx-apidoc, documentation from private modules were being included via python domain directives. The 2 occurrences of this include:
- The [common](https://github.com/ansys/pydpf-core/blob/d5ed8ece6b31b4e5445fdc4fd790c61d66404114/src/ansys/dpf/core/common.py#L27) module.
- The [error](https://github.com/ansys/pydpf-core/blob/d5ed8ece6b31b4e5445fdc4fd790c61d66404114/src/ansys/dpf/core/errors.py#L29) module.

The above logic works because apidoc executes imports, so it is still able to resolve the location of the documented python objects even though the */gate/ folder is [excluded](https://github.com/ansys/pydpf-core/blob/d5ed8ece6b31b4e5445fdc4fd790c61d66404114/.ci/build_doc.bat#L7) from the api documentation.

Due to switching to autoapi, this logic will not work when using [equivalent autoapi directives](https://sphinx-autoapi.readthedocs.io/en/latest/reference/directives.html#autodoc-style-directives) because autoapi parses files and never executes imports, making it unable to resolve the location of the python objects.

This issue should address the best way forward to solve this problem.

PProfizi:
I think we should move the referenced objects to the core sub-package. The `locations` HAVE to be documented and available publicly as it is a cornerstone of how DPF works.
As for error, I think error types should be public too and documented as we then would be able to add explanations and examples to their docstrings.

PR #2375, opened 2025-06-17
fix(api): Make locations a publicly documented class

PProfizi:
Answers part of #2009 

Until now the `locations` enum has not been available in the API reference since it is defined in `ansys/dpf/gate/common` which is excluded from the API reference.

This is very problematic as we reference locations all the time in the rest of the doc but the Enum in itself turns out to be hidden.


rafacanton (approved):

PProfizi:
> @PProfizi I approved but saw that post is broken, I guess that a PR is needed there following this one

Yes, I will adapt Post right away, thanks!

PR #2386, opened 2025-06-20
feat(server): Add ServerContext.xml_path setter

PProfizi:
To allow setting the `xml_path` for a `dpf.AvailableServerContexts.custom_defined` context.

First remark: the right `AvailableServerContext` for using a custom xml path is `no_context` (`context_type=2`) instead of `custom_defined` (`context_type=3`)... this is confusing.

Second remark: despite the tests passing on my local machine with a unified Ansys install, it does not pass in the GH pipelines with a standalone DPF. Is there something I am missing?

@rafacanton @ansys-akarcher I think I will need some help on this one. This comes from a workaround shared to a client where they needed to remove `multiphysics_mapper` from default plugins. It is easier for them to share a custom `XML` file and use it within a PyDPF script rather than modifying their actual DPF installation.
The workaround shared was to use `ServerContext(context_type=2, xml_path="path/to/file")`.
The point here is to make this user-friendly.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -173,6 +173,7 @@ class TestServer:
 @pytest.mark.skipif(
     not SERVERS_VERSION_GREATER_THAN_OR_EQUAL_TO_10_0, reason="not working properly before 25R2"
 )
+@pytest.mark.skipif(running_docker, reason="server start using custom xml not working on Docker")
 def test_server_context_custom_xml(remote_config_server_type, testfiles_dir):
     from pathlib import Path
 
@@ -180,24 +181,6 @@ def test_server_context_custom_xml(remote_config_server_type, testfiles_dir):
     context.xml_path = Path(testfiles_dir) / "DpfCustomDefinedTest.xml"
     server_plugins = start_local_server(config=remote_config_server_type, context=context).plugins
     ref = ["grpc", "native"]
-    if running_docker:  # Use of custom xml not working either on Docker
-        ref = [
-            "cff",
-            "cngs",
-            "compression",
-            "documentation",
-            "fem_utils",
-            "grpc",
-            "hdf5",
-            "live_post",
-            "lsDyna",
-            "mapdl_plugin",
-            "math",
-            "mechanical_results",
-            "mesh_plugin",
-            "native",
-            "vtk",
-        ]
     assert sorted(list(server_plugins.keys())) == ref
 
 


tests/test_server.py
```diff
@@ -180,6 +181,60 @@ def test_server_plugins(self, server_config):
...
+            "mechanical",
+            "mechanical_results",
+            "mesh_operators",
+            "multiphysics_mapper",
+            "native",
+            "point_cloud_search",
+            "prime",
+            "python_loader",
+            "rbd",
+            "vtk",
```
rafacanton, line 233: If you remove `point_cloud_search`, `multiphysics_mapper` and `python_loader` from this list, I think the tests will pass

tests/test_server.py
```diff
@@ -180,6 +181,60 @@ def test_server_plugins(self, server_config):
         assert "native" in server_plugins.keys()
 
 
+@raises_for_servers_version_under("7.0")
```
PProfizi, line 184:
```suggestion
@pytest.mark.skipif(not SERVERS_VERSION_GREATER_THAN_OR_EQUAL_TO_10_0, reason="not working properly before 25R2")
```

tests/test_server.py
```diff
@@ -180,6 +181,60 @@ def test_server_plugins(self, server_config):
...
+def test_server_context_custom_xml(remote_config_server_type, testfiles_dir):
+    from pathlib import Path
+
+    context = dpf.core.AvailableServerContexts.no_context
+    context.xml_path = Path(testfiles_dir) / "DpfCustomDefinedTest.xml"
+    server_plugins = start_local_server(config=remote_config_server_type, context=context).plugins
+    if SERVERS_VERSION_GREATER_THAN_OR_EQUAL_TO_10_0:
+        ref = ["grpc", "native"]
+    elif running_docker:  # Use of custom xml not working either on Docker
```
PProfizi, line 193:
```suggestion
    ref = ["grpc", "native"]
    if running_docker:  # Use of custom xml not working either on Docker
```

tests/test_server.py
```diff
@@ -169,6 +170,37 @@ def test_server_plugins(self, server_config):
...
+    not SERVERS_VERSION_GREATER_THAN_OR_EQUAL_TO_10_0, reason="not working properly before 25R2"
+)
+def test_server_context_custom_xml(remote_config_server_type, testfiles_dir):
+    from pathlib import Path
+
+    context = dpf.core.AvailableServerContexts.no_context
+    context.xml_path = Path(testfiles_dir) / "DpfCustomDefinedTest.xml"
+    server_plugins = start_local_server(config=remote_config_server_type, context=context).plugins
+    ref = ["grpc", "native"]
+    if running_docker:  # Use of custom xml not working either on Docker
```
ansjhajjar: If it does not work, maybe we should skip it in that case ?

tests/test_server.py
```diff
@@ -169,6 +170,37 @@ def test_server_plugins(self, server_config):
         assert "native" in server_plugins.keys()
 
 
+@pytest.mark.skipif(
+    not SERVERS_VERSION_GREATER_THAN_OR_EQUAL_TO_10_0, reason="not working properly before 25R2"
+)
+def test_server_context_custom_xml(remote_config_server_type, testfiles_dir):
```
PProfizi:
```suggestion
@pytest.mark.skipif(running_docker, reason="server start using custom xml not working on Docker")
def test_server_context_custom_xml(remote_config_server_type, testfiles_dir):
```

PR #2439, opened 2025-07-16
docs: fix image link in README.md

SMoraisAnsys:
I would recommend to use `main` as default branch. In the meantime, the doc can be fixed by pointing to the current default (`master`).


PProfizi:
Default branch is now `main`. Closing this issue. Thanks @SMoraisAnsys!

PR #2469, opened 2025-07-24
Disable CFF test (temporary)

rafacanton:
This test needs to be temporarily disabled to ensure pipeline stability until servers are updated to be able to readd all fluid variables constained in CFF files.

https://github.com/ansys/pydpf-core/pull/2468 has the actual changes when the server is up-to-date

### DIFF
--- a/tests/test_resultinfo.py
+++ b/tests/test_resultinfo.py
@@ -201,52 +201,52 @@ Available qualifier labels:"""  # noqa: E501
         assert len(ar.qualifier_combinations) == 20
 
 
-@pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


PProfizi:
Hey @rafacanton could we actually instead make the tests better to start with, like was done in #2436 for MAPDL result info?

rafacanton:
> Hey @rafacanton could we actually instead make the tests better to start with, like was done in #2436 for MAPDL result info? The idea being that we stop having to do this kind of manipulation. The problem here is test design. I can take the ownership of the PR and do it if you want.

Yes, see the updated test here: https://github.com/ansys/pydpf-core/pull/2468

PProfizi (approved):

PR #2529, opened 2025-08-07
ci(doc): Switch doc-deploy-pr to a dedicated workflow

PProfizi:
Moves the `doc-deploy-pr` action to a dedicated workflow triggered by a `workflow_run` `completed` on `docs.yml`, and deploying only if the PR is labeled with `deploy-pr-doc`.
The new workflow needs to be present in the main branch to be triggered.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review, closed, labeled]
+    types: [opened, synchronize, reopened, ready_for_review]
     branches-ignore:
       - '*no-ci*'
   push:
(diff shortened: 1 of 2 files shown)


moe-ad:
> > @PProfizi nice approach. I didn't even know an action exist for downloading artifacts from other workflows. I made some suggestions, especially wrt the need for the closed event to be present.
> > I still have a concern: I think this shouldn't need to be on main before we can confirm that it works because reusable workflows are picked up from the current branch, the reason it currently doesn't get triggered might be because we run `docs.yml` from `ci.yml`, and it may not be considered valid for the `workflow_run` trigger. I may be wrong.
> 
> According to the doc you can trigger up to 3 workflows in a sequence this way

Hi @PProfizi, I just saw the section in the docs where it was stated that the workflow [has to be on the default branch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows?versionId=free-pro-team%40latest&productId=actions&restPage=reference%2Cworkflows-and-actions%2Cworkflow-syntax#workflow_run) for this to work.. :)

PR #2536, opened 2025-08-08
ci: update deploy_pr_doc logic

PProfizi:
Based on @moe-ad's comments in #2529


PProfizi:
@moe-ad it looks like it is still not triggered

Issue #2491, opened 2025-07-29
dpf issues DepreciationWarnings against itself


andrew-d4:
### Before submitting the issue

- [x] I have checked for [Compatibility](https://dpf.docs.pyansys.com/version/stable/getting_started/compatibility.html#id1) issues
- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

This isn't a bug but it is annoying.

When running tests on code that uses dpf, I am getting multiple DepreciationWarnings which are issued by the following line:

`Lib\site-packages\ansys\dpf\core\inputs.py:240: DeprecationWarning: Use explicit output-to-input connections.`

If I look at the stack trace when these warnings are generated they are caused by dpf code calling the .connect method.

The first 2 examples I found were:
`dpf.core.model.Metadata.mesh_provider` when running the line `mesh_provider.inputs.connect(self._stream_provider.outputs)`

`dpf.core.model.Metadata._cache_streams_provider` when running the line `self._stream_provider.inputs.connect(self._data_sources)`

### Steps To Reproduce

Run pytest on some code that access the functions above. 

### Which Operating System causes the issue?

Windows

### Which DPF/Ansys version are you using?

Ansys 2024 R2

### Which Python version causes the issue?

3.11

### Installed packages

ansys-api-mapdl==0.5.2
ansys-api-platform-instancemanagement==1.1.3
ansys-dpf-core==0.13.8
ansys-mapdl-core==0.68.4
ansys-mapdl-reader==0.54.2
ansys-math-core==0.1.5
ansys-platform-instancemanagement==1.1.2
ansys-tools-path==0.7.1

PProfizi:
Hello @andrew-d4,

Thanks for raising this, it is indeed a bug if we still have some lines using that connection logic.
I will look into refactoring them.

PR #2551, opened 2025-08-19
chore: remove all deprecated calls in pydpf-core

PProfizi:
Solves #2491 

Enable all warnings during testing
Then fix all calls to deprecated methods such as `Inputs.connect` and `Outputs.connect`.


src/ansys/dpf/core/model.py
```diff
@@ -31,6 +31,7 @@
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:  # pragma: nocover
+    from ansys.dpf.core.operators.mesh.mesh_provider import mesh_provider as mesh_provider_op
```
BClappe, line 34: Is it really what works, or should it be what's below?
```suggestion
    from ansys.dpf.core.operators.mesh import mesh_provider as mesh_provider_op
```

But I guess that if everything passes, it's probably what you meant to do

BClappe (approved):

PR #2603, opened 2025-09-15
Replace displacement operator by modal coordinate

oparreno:


oparreno:
@rafacanton, @PProfizi, would you please approve and merge this PR? Thanks

rafacanton:
@oparreno As `modal_coordinate` just exists for the 26.1 server, we need to ensure that a user with the latest pydpf and an old server is able to run the operator and the example. If you add it here: https://github.com/ansys/pydpf-core/blob/47115019857f5cd664347036291fd252c3e9b43f/src/ansys/dpf/core/operators/build.py#L21, where the key is the new operator "MCF" and the value is the operator to redirect in old servers "U".

PR #2646, opened 2025-10-01
fix: keep import of ServerBase in server.py

PProfizi:
This was removed during of modification of typing but should be kept for backward compatibility of imports.

### DIFF
--- a/src/ansys/dpf/core/server.py
+++ b/src/ansys/dpf/core/server.py
@@ -48,7 +48,13 @@ from ansys.dpf.core.server_factory import (
     ServerConfig,
     ServerFactory,
 )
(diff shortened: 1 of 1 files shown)


src/ansys/dpf/core/server.py
```diff
@@ -48,7 +48,13 @@
     ServerConfig,
     ServerFactory,
 )
-from ansys.dpf.core.server_types import DPF_DEFAULT_PORT, LOCALHOST, RUNNING_DOCKER, AnyServerType
+from ansys.dpf.core.server_types import (
+    DPF_DEFAULT_PORT, 
+    LOCALHOST, 
+    RUNNING_DOCKER, 
+    ServerBase,
```
PProfizi, line 55:
```suggestion
    BaseServer, 
```

cbellot000 (approved):

PR #2661, opened 2025-10-08
feat: give a default name to loaded plugins

PProfizi:
The default name given to plugins loaded using `load_library` is an empty string and not a `None`, which creates a name conflict if two plugins are loaded without explicit names.
This PR adds a default name to plugins loaded with no `name` argument.


src/ansys/dpf/core/core.py
```diff
@@ -412,7 +430,13 @@ def load_library(self, file_path, name="", symbol="LoadOperators", generate_oper
         >>> # base.load_library('meshOperatorsCore.dll', 'mesh_operators')
 
         """
-        file_path = str(file_path)
+        if not name:
```
PProfizi, line 433:
```suggestion
        if name is None:
```
@rafacanton the None is never passed to the server, so there is no expected error.
The server API only accepts a string, which is why we used to pass an empty string.

tests/test_plugins.py
```diff
@@ -121,3 +122,18 @@ def test_vtk(server_type, tmpdir):
...
+
+
+@pytest.mark.xfail(raises=errors.DpfVersionNotSupported)
+def test_load_library_default_name(remote_config_server_type):
+    # Test only for remote server configs as InProcess already ran and loaded plugins at this point
+    xml_path = Path(conftest.DEFAULT_ANSYS_PATH) / "dpf" / "utilities" / "DpfCustomDefined.xml"
+    server_context = dpf.server_context.ServerContext(context_type=2, xml_path=str(xml_path))
+    print(server_context)
+    server = dpf.start_local_server(
+        config=config_server_type, context=server_context, as_global=False
```
PProfizi:
```suggestion
        config=remote_config_server_type, context=server_context, as_global=False
```

rafacanton (approved):

PR #2695, opened 2025-10-23
fix: design and type hint of custom collections

### LINKED ISSUE #2288: Custom collections do not show any available methods in IDE

### Description of the bug

A `GenericDataContainersCollection` does not show any available methods in IDEs when trying autocompletion.
Probably due to the design for custom collections.
This is a major issue with this design.

### Steps To Reproduce

try to get available methods with:
```py
bc = dpf.GenericDataContainersCollection()
bc.  # -> should propose methods from Collection
```

Windows

DPF Server 2025.2.pre0

3.12

### Installed packages

NA

moe-ad:
Closes #2288.

The following changes have been made:
- Making the fact that `Collection` is also a generic type more explicit. From my investigation, this helps older type checker versions (especially on older python versions) infer types more accurately.
- Make the `CollectionFactory` function a `collection_factory` class method of `Collection`. This is done for three main reasons:
  - It is makes more sense and it is more intuitive to bind it to the class, making it more obvious that it returns class objects deriving from `Collection`. 
  - Eliminating the need to pass an explicit `BaseClass` argument. I couldn't think of a reason why a different argument other than `Collection` will be passed. @PProfizi what do you think?
  - To accurately type-hint the factory function, access to the type parameter of `Collection` is needed.
- Corrections and improvements to the type hints.

With these changes, IDEs are now better able to suggest accurate autocompletion. Example in PyCharm:
image

The type inference is also much more accurate. For example, IDEs are able to infer that a `StringFieldsCollection` contains `StringFields`:
image

And because of that, the type of the object returned by a call to `get_entry` will be inferred correctly as `StringField`:
image

And only available methods of that object will be suggested through autocompletion:
image

### DIFF
--- a/src/ansys/dpf/core/documentation/generate_operators_doc.py
+++ b/src/ansys/dpf/core/documentation/generate_operators_doc.py
@@ -120,7 +120,9 @@ def extract_operator_description_update(content: str) -> str:
             The updated description to use for the operator.
     """
     match = re.search(r"## Description\s*(.*?)\s*(?=## |\Z)", content, re.DOTALL)
-    return match.group(0) + os.linesep if match else None
+    description = match.group(0) + os.linesep if match else None
+    # Handle unicode characters
+    return description.encode("unicode-escape").decode()
 
 
 def replace_operator_description(original_documentation: str, new_description: str):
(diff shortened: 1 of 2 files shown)


src/ansys/dpf/core/collection.py
```diff
@@ -25,13 +25,17 @@
...
 
+from typing import Generic, Type, TypeVar
+
 from ansys.dpf.core import errors, server as server_module
 from ansys.dpf.core.any import Any
-from ansys.dpf.core.collection_base import TYPE, CollectionBase
+from ansys.dpf.core.collection_base import CollectionBase
 from ansys.dpf.core.common import create_dpf_instance
 
+TYPE = TypeVar("TYPE")
```
jorgepiloto: Let's use `T` as per the original [PEP 484]:


```suggestion
T = TypeVar("T")
```

[PEP 484]: https://peps.python.org/pep-0484/#user-defined-generic-types

PR #2732, opened 2025-11-06
feat(typing): improve type hinting of `Input`/`Output` classes and `operators` subpackage

### LINKED ISSUE #2701: Properly type operator inputs/outputs

### Description of the bug

Typing of operator pins is not working properly.

Taking `stress` as an example, no typing is generated for the [constructor](https://github.com/ansys/pydpf-core/blob/613e2205f4337d84a1ad4ea736439eca5f9e8c70/src/ansys/dpf/core/operators/result/stress.py#L239), and 
no type deduction or propagation is done when creating the `Input` or `Output` instance for each input/output of the operator (see for example [here](https://github.com/ansys/pydpf-core/blob/613e2205f4337d84a1ad4ea736439eca5f9e8c70/src/ansys/dpf/core/operators/result/stress.py#L737)).
Everything is based on the typing of the base [Input.connect](https://github.com/ansys/pydpf-core/blob/613e2205f4337d84a1ad4ea736439eca5f9e8c70/src/ansys/dpf/core/inputs.py#L78) or [Output.get_data](https://github.com/ansys/pydpf-core/blob/613e2205f4337d84a1ad4ea736439eca5f9e8c70/src/ansys/dpf/core/outputs.py#L59).

Operator code generation is handled [here](https://github.com/ansys/pydpf-core/blob/613e2205f4337d84a1ad4ea736439eca5f9e8c70/src/ansys/dpf/core/operators/build.py#L187).

### Steps To Reproduce

Here is a simple example where we connect a `DataSources` to the `data_sources` input pin of an operator.
My IDE marks `ds` as of unexpected type in the call to `connect` in `s_op.inputs.data_sources.connect(ds)`.
In the same manner, one would expect `s_out` type to be understood as `FieldsContainer` or `dpf.types.fields_container` but it is instead marked as `Any`.

```py
import ansys.dpf.core as dpf


ds = dpf.DataSources(examples.download_static_rst())

s_op = dpf.operators.result.stress()
s_op.inputs.data_sources.connect(ds)  # Should accept a DataSources type
s_out = s_op.outputs.fields_container()  # Should return a FieldsContainer type
```

Windows

Ansys 2026 R1

3.11

### Installed packages

NA

moe-ad:
Closes #2701.


src/ansys/dpf/core/operators/averaging/elemental_difference.py
```diff
@@ -14,6 +14,12 @@
 from ansys.dpf.core.config import Config
 from ansys.dpf.core.server_types import AnyServerType
 
+# For type checking
+from ansys.dpf.core.field import Field
```
PProfizi: @moe-ad could we put that in an `if TYPE_CHECKING` using `from typing import TYPE_CHECKING` just to make sure it will not create import issues down the line?

src/ansys/dpf/core/operators/__init__.py
```diff
@@ -2,7 +2,6 @@
 from . import compression
 from . import filter
 from . import geo
-from . import info
```
PProfizi: Hi @moe-ad it looks like you ran the operator update locally and do not have the most recent changes server-side. You'll need to run an update directly on GitHub targeting your branch.

PR #2777, opened 2025-11-24
feat(operator_doc): include operator usage examples in the operator documentation

PProfizi:
This PR updates the `ansys.dpf.core.documentation` package to integrate script usage examples for C++, IPython and CPython in the operator documentation for the DPF Framework documentation.


PProfizi:
Here is an example of how they render when I build the DPF Framework doc for 26R1 locally:

image

PR #2790, opened 2025-12-01
grpc changes

ansys-akarcher:


RobPasMue (changes requested):
Would it be possible to use the ansys.tools.common.cyberchannel module inside the ansys-tools-common package rather than a local copy of the cyberchannel module? We are pushing all teams for this request

src/ansys/dpf/core/operators/scoping/adapt_with_scopings_container.py
```diff
@@ -118,6 +118,12 @@ def _spec() -> Specification:
                     optional=True,
                     document=r"""Whether to keep fields that become empty after rescoping. Default is false.""",
                 ),
+                2: PinSpecification(
```
rafacanton: @ansys-akarcher what do the changes in this operator and in the following one have to do with the objective of the PR?

src/ansys/dpf/core/server_types.py
```diff
@@ -1229,11 +1275,18 @@ def __init__(
         docker_config: DockerConfig = RUNNING_DOCKER,
         use_pypim: bool = True,
         context: server_context.ServerContext = server_context.SERVER_CONTEXT,
+        grpc_mode: server_factory.GrpcMode = server_factory.DEFAULT_GRPC_MODE,
```
PProfizi: @ansys-akarcher do we really need `mTLS` for InProcess? What's the potential risk? There is no gRPC connection here.

src/ansys/dpf/core/cyberchannel.py
```diff
@@ -0,0 +1,514 @@
...
+"""
+
+# Only the create_channel function is exposed for external use
+__all__ = ["create_channel", "verify_transport_mode", "verify_uds_socket"]
+
+from dataclasses import dataclass
+import logging
+import os
+from pathlib import Path
+from typing import TypeGuard, cast
```
rafacanton:
```suggestion
from typing import cast
```

RobPasMue:
@ansys-akarcher @rafacanton @PProfizi - are we planning on releasing these changes shortly? Just to make them available to end users as needed for the older releases.

PR #2823, opened 2025-12-10
fix test_get_resultinfo_2

YounMars:


rafacanton:
@YounMars This will not respect retro. Please add a condition to look for FULL in old releases and MAPDL inlatest. You can see similar conditions in other tests

PR #2880, opened 2026-01-05
refactor(grpc): switch to using ansys-tools-common's cyberchannel module

### LINKED ISSUE #2875: Using ansys-tools-common package for instantiating gRPC channels
> Would it be possible to use the ansys.tools.common.cyberchannel module inside the ansys-tools-common package rather than a local copy of the cyberchannel module? We are pushing all teams for this request

@RobPasMue we are going to merge a first version of this PR as soon as possible to allow for changes server-side, but later work Python side will include using the `ansys-tools-common` library as well as writing all the necessary documentation.

_Originally posted by @PProfizi in https://github.com/ansys/pydpf-core/issues/2790#issuecomment-3641261792_

PProfizi:
Closes #2875

### DIFF
--- a/requirements/requirements_test.txt
+++ b/requirements/requirements_test.txt
@@ -1,4 +1,5 @@
 ansys-platform-instancemanagement==1.1.2
+ansys-tools-common==0.4.0
 coverage==7.13.0
 graphviz==0.21
 imageio==2.37.0


RobPasMue (approved):
LGTM! Thanks @PProfizi !

PProfizi:
I noticed that this PR forces us to explicitly drop support for Python 3.9, which we had not done yet.

PR #2881, opened 2026-01-05
feat: support viz interface

### LINKED ISSUE #2841: Update plotting modules to support `ansys-tools-visualization-interface`.
### Value proposition
The main value proposition is that [`ansys-tools-visualization-interface`](https://visualization-interface.tools.docs.pyansys.com/version/stable/index.html) has some level support for other backends (e.g. [`trame`](https://kitware.github.io/trame/) and [`plotly`](https://plotly.com/)) in addition to support for `pyvista`(which `pydpf-core` currently leverages). Which means we could also support more plotting backends other than pyvista through the viz-interface.

I would also like to mention is that other flagship libraries have integrated the viz-interface as well, and doing same in this library would mean we now expose plotting functionalities that are similar in look and feel to other PyAnsys  ecosystem libraries.

Furthermore, new features such as OpenUSD support are in the pipelines for the viz-interface that we could benefit from (and of course any new features in the future as well).

We can start exploring how best to integrate the viz interface in `pydpf-core` as well.

### General notes
While we are still refining the implementation approach continuously, the following decisions have been reached so far:
- The plan is not to expose the same plotting APIs as other libraries using the viz-interface. No common API currently exists and we are free to customize things as needed.
- For starters, we should focus on switching the current [`_PyVistaPlotter`](https://github.com/ansys/pydpf-core/blob/260687281301e2aea50f9bb964b3eb6345ab1ef3/src/ansys/dpf/core/plotter.py#L63) backend to viz-interface's `PyVistaBackend`.

moe-ad:
Closes #2841.

#### Summary of changes
- To prevent direct coupling between the plotter that has to be implemented and viz-interface's plotter backends, we had to propose a couple of backend agnostic customization APIs in the viz-interface. For more context regarding that, see:
  - ansys/ansys-tools-visualization-interface#448
  - ansys/ansys-tools-visualization-interface#465
  - ansys/ansys-tools-visualization-interface#466
  - ansys/ansys-tools-visualization-interface#474
- A `_VisualizationInterfacePlotter` class has been added. Since viz's `PlotlyBackend` doesn't support volumetric meshes like the `PyVistaBackend` does, we are plugging `pydpf-core`'s plotter to only the latter.
- We still can't deprecate the now legacy `_PyVistaPlotter` yet, since other classes like `Plotter` and `Animator` depends on it. Once those are gradually phased out via separate PRs in the future, we will eventually be able to deprecate `_PyVistaPlotter`.
- Because of the previous point, the plotter factory has been parametrized with a `plotter_type` argument that can be one of `PlotterBackend.VISUALIZATION_INTERFACE`, `PlotterBackend.PYVISTA`, and `PlotterBackend.AUTO`. The behavior is as follows:
  -  `PlotterBackend.AUTO` is the default. This will select the new plotter or the legacy plotter depending on which one is available in the user's environment. This guarantees that plotting scripts won't automatically start breaking if users update `pydpf-core` without updating it's plotting dependencies.
  - `PlotterBackend.PYVISTA` will select the legacy plotter and `PlotterBackend.VISUALIZATION_INTERFACE` will select the new plotter.

@PProfizi and @jorgepiloto, this is now ready for review.

> [!NOTE]
> We will address switching to the viz-interface as the default backend in a separate PR. It requires making sure all tests pass and that might require a bit more work.


PProfizi:
Hi @moe-ad the example for node labels does not work 
https://dpf.docs.pyansys.com/pull/2881/examples/06-plotting/03-labels.html
Here is the reference
https://dpf.docs.pyansys.com/version/dev/examples/06-plotting/03-labels.html

PProfizi (approved):

PR #2957, opened 2026-02-03
fix: remove forced debug logs

### LINKED ISSUE #2928: Please allow clean logging: do not set log level in local loggers

### Description of the bug

In several modules a local logger `LOG` is defined (ok and idiomatic), but its level is set to DEBUG, and not left unset. This means that unwanted DEBUG log messages pop up from time to time in scripts using pydpf-core.

The actual problem arises in 

https://github.com/ansys/pydpf-core/blob/349f0fddb28473278f1b8b5139797048a9faf658/src/ansys/dpf/core/server_types.py#L62-L63

In this module there are a lot of calls to `LOG.debug` which surface even if the root logger is set to a higher level. Fix is simple to delete/comment out line 63 in production releases.

There are other instances which do not cause problem because `LOG` is not used in the module (please consider to delete the unused variables).

https://github.com/ansys/pydpf-core/blob/349f0fddb28473278f1b8b5139797048a9faf658/src/ansys/dpf/core/core.py#L61-L62
https://github.com/ansys/pydpf-core/blob/349f0fddb28473278f1b8b5139797048a9faf658/src/ansys/dpf/core/dpf_operator.py#L68-L69
https://github.com/ansys/pydpf-core/blob/349f0fddb28473278f1b8b5139797048a9faf658/src/ansys/dpf/core/session.py#L41-L42
https://github.com/ansys/pydpf-core/blob/349f0fddb28473278f1b8b5139797048a9faf658/src/ansys/dpf/core/workflow.py#L56-L57

It seems to me that this is simply development code which has slipped into production releases.

### Steps To Reproduce

Simply run
```
>>> import logging
>>> logging.basicConfig(level=logging.ERROR)
>>> server = dpf.server.connect_to_server()
```
The following log message pops up, which should instead be filtered by the root logger which is set at a higher level.
```
DEBUG:ansys.dpf.core.server_types:Established connection to DPF gRPC
```

MacOS

DPF Server 2025.1.pre0

3.13

### Installed packages

```
ansys-tools-common==0.4.1
certifi==2026.1.4
charset-normalizer==3.4.4
click==8.3.1
grpcio==1.76.0
idna==3.11
importlib-metadata==8.7.1
numpy==2.4.1
packaging==26.0
platformdirs==4.5.1
protobuf==6.33.4
psutil==7.2.1
requests==2.32.5
scooby==0.11.0
setuptools==80.10.2
tqdm==4.67.1
typing-extensions==4.15.0
urllib3==2.6.3
zipp==3.23.0
```

PProfizi:
Closes #2928

### DIFF
--- a/src/ansys/dpf/core/session.py
+++ b/src/ansys/dpf/core/session.py
@@ -38,6 +38,8 @@ from ansys.dpf.core.common import (
 )
 from ansys.dpf.gate import capi, session_capi, session_grpcapi
 
+LOG = logging.getLogger(__name__)
+
 
 @capi.GenericCallBackType
 def progress_call_back(obj, nature, arg):


moe-ad (approved):
LGTM!

PR #2970, opened 2026-02-09
test: Small change to energy units in Imperial

rafacanton:
React to the server changes to further improve energy units of Imperial systems from `lbf*ft` and `lbf*in` to `ft*lbf` and `in*lbf`.


rafacanton:
Failure expected until the server is updated

PR #2990, opened 2026-02-17
Avoid misc gate client config mock class

ansys-akarcher:
Temporarily passing the LocalClientConfig was wrongly initializing RuntimeclientConfig, and on the GrpcClient side, it would add grpc a metadata interceptor, resulting in an unexpected performance regression on all grpc requests.

### DIFF
--- a/src/ansys/dpf/core/server_types.py
+++ b/src/ansys/dpf/core/server_types.py
@@ -923,7 +923,6 @@ class GrpcServer(CServer):
                 )
                 self._local_server = True
 
-        from ansys.dpf.core import settings
         client_config = settings.get_runtime_client_config(server=self)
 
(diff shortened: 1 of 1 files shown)


src/ansys/dpf/core/server_types.py
```diff
@@ -923,13 +923,15 @@ def __init__(
                 )
                 self._local_server = True
 
-        local_client_config = settings.get_runtime_client_config()
+        from ansys.dpf.core import settings
+        client_config = settings.get_runtime_client_config(server=self)
```
PProfizi: @ansys-akarcher so if I understand correctly, the fix is to pass `server=self` here?

rafacanton:
@PProfizi This looks good, we will need your bypassing rights for the coverage (as secure is not really covered in tests...). FYI @ansys-akarcher

Issue #3004, opened 2026-02-24
`Field().deep_copy()` and `FieldsContainer().deep_copy()` don't work with dimensionless units


ansaminard:
### Before submitting the issue

- [x] I have checked for [Compatibility](https://dpf.docs.pyansys.com/version/stable/getting_started/compatibility.html#id1) issues
- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

Invalid unit errors are raised when attempting to create deep copies of fields and fields container with dimensionless units, even if the units were properly set as `tuple[Homogeneity, str]`

For example:
```
# Set a field's unit as dimensionless
my_field.unit = (Homogeneity.dimensionless, "some_unit")
# Attempt to create a deep copy
my_field_copy = my_field.deep_copy()
```
or
```
# Set a fields container's unit as dimensionless
for field in my_fields_container:
    field.unit = (Homogeneity.dimensionless, "some_unit")
# Attempt to create a deep copy
my_fields_container_copy = my_fields_container.deep_copy()
```
both raise the following error:
`ansys.dpf.gate.errors.DPFServerException: dataProcessing::FieldDefinition_SetUnit: Invalid argument: "dBSPL" is not a valid unit.`

More generally, if you use the `unit` property, previously set as dimensionless, of an existing field into another field directly, you'll get the error as well.

On a sidenote, it would be useful to be able to get, in PyDPF Core, the list of supported (ie non-dimensionless) units, to be able to determine whether a given unit must be set as dimensionless or not.

### Steps To Reproduce

```
# Set a field's unit as dimensionless
my_field.unit = (Homogeneity.dimensionless, "some_unit")
# Attempt to create a deep copy
my_field_copy = my_field.deep_copy()
```

### Which Operating System causes the issue?

Linux, Windows

### Which DPF/Ansys version are you using?

Ansys 2026 R1

### Which Python version causes the issue?

3.13

### Installed packages

ansys-dpf-core==0.15.0
-e git+https://github.com/ansys/pyansys-sound.git@9dd41bbe8bb47487469a9968a2ec2456400902a5#egg=ansys_sound_core
ansys-tools-common==0.4.0
appdirs==1.4.4
certifi==2026.1.4
cfgv==3.5.0
charset-normalizer==3.4.4
click==8.3.1
colorama==0.4.6
contourpy==1.3.3
coverage==7.13.1
cycler==0.12.1
distlib==0.4.0
filelock==3.20.3
[...]

PR #3007, opened 2026-02-25
fix: issue with deep copy units with homogeneity dimensionless

jvidalor:
Updating code to support deep_copy() for fields (related to this [issue ](https://github.com/ansys/pydpf-core/issues/3004)).
We are now using getting the homogeneity with the units, if it is dimensionless.

### DIFF
--- a/src/ansys/dpf/core/field_definition.py
+++ b/src/ansys/dpf/core/field_definition.py
@@ -118,7 +118,7 @@ class FieldDefinition:
         -------
         str or tuple
             Units of the field. If the field has a dimensionless homogeneity with a named unit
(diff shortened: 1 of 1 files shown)


src/ansys/dpf/core/field_definition.py
```diff
@@ -112,18 +112,29 @@ def unit(self):
 
         Returns
         -------
-        str
-            Units of the field.
+        str or tuple
```
PProfizi: Hi @jvidalor, could we instead provide a `FieldDefinition.homogeneity` getter?

rafacanton:
@jvidalor can you change the PR name to put it in Conventional Commits format?

PR #3155, opened 2026-04-28
fix: deep_copy of FieldsContainers with GenericSupports

rafacanton:
Fixes #3041


rafacanton:
Tests are currently failing pending on the server update

src/ansys/dpf/core/any.py
```diff
@@ -211,6 +213,16 @@ def _type_to_new_from_get_as_method(self, obj):
                 self._api.any_new_from_data_sources,
                 self._api.any_get_as_data_sources,
             )
+        elif issubclass(obj, generic_support.GenericSupport):
```
PProfizi, line 216: @rafacanton just a thought:
maybe we should store the `any_new_from` and `any_get_as` api methods directly in each class as static properties or something so we do not have to maintain this long if chain.
We could then simply get them from the type.

src/ansys/dpf/core/collection_base.py
```diff
@@ -534,7 +535,25 @@ def get_support(self, label: str) -> Support:
         """
         from ansys.dpf.core.support import Support
 
-        return Support(support=self._api.collection_get_support(self, label), server=self._server)
+        support = self._api.collection_get_support(self, label)
+        if support:
+            return Support(support=support, server=self._server)
+        else:
+            return support
```
PProfizi, line 542: @rafacanton that means the method can return `None`, right?

PProfizi: Or should we return an empty support as I guess it was done previously.

rafacanton: correct, if there is not a support for that label. For example, a FC on "complex" and "time", if you currently call this method with the label "complex", the server will return nullptr. We were previously not making any check at all, and the resulting Support had no _internal_object. Now if the label has no support, we return None, which I think is more explicit

src/ansys/dpf/core/support.py
```diff
@@ -77,7 +84,27 @@ def __init__(self, support, server=None):
 
         # step4: take instance:
         # object_name -> protobuf.message, DPFObject*
-        self._internal_obj = support
+        if support is not None:
+            from ansys.dpf.core import CyclicSupport, GenericSupport, MeshedRegion, TimeFreqSupport
+
+            if isinstance(
+                support, (Support, TimeFreqSupport, MeshedRegion, GenericSupport, CyclicSupport)
```
PProfizi, line 91: @rafacanton so we are going to have to maintain this list. Can't we use subclass detection?

src/ansys/dpf/core/support.py
```diff
@@ -191,6 +218,124 @@ def available_string_field_supported_properties(self):
...
+        from ansys.dpf.core import Any, CyclicSupport, GenericSupport
+
+        support_type = self.get_type()
+        if support_type == "TimeFreqSupport":
+            return Support(support=self.get_as_time_freq_support().deep_copy(server), server=server)
+        elif support_type == "CMeshDomainSupport":
+            return Support(support=self.get_as_meshed_region().deep_copy(server), server=server)
+        elif support_type == "cyclic_support":
+            return Support(
+                support=_deep_copy(
```
PProfizi, line 326: @rafacanton can we maybe explain why this one is different?

PProfizi: I guess because it does not have a `deep_copy` method?

src/ansys/dpf/core/support.py
```diff
@@ -191,6 +218,124 @@ def available_string_field_supported_properties(self):
...
+            global server.
+
+        Returns
+        -------
+        support : Support
+        """
+        from ansys.dpf.core import Any, CyclicSupport, GenericSupport
+
+        support_type = self.get_type()
+        if support_type == "TimeFreqSupport":
```
PProfizi, line 320: @rafacanton again this is a list of IFs, maybe we could refactor this?

src/ansys/dpf/core/support.py
rafacanton: @PProfizi correct

src/ansys/dpf/core/support.py
rafacanton: @PProfizi The problem is that GenericSupport inherits from Support, but not the rest. I guess because they were added later on, and I cannot calibrate the consequences of making them inherit from Support right now. That's why we have all the ifs. I have tried to only have them inside the Support class, so the users of the Support are unaware of these internal details

src/ansys/dpf/core/support.py
rafacanton: Same answer as before. Once I have a support, I don't really know its type and how to deep_copy, hence I ask the server to tell it to me

src/ansys/dpf/core/any.py
rafacanton: Yes, I think this would be great. I can add a follow-up issue to change this, as it is beyond of the objective of this bug fix

PProfizi: That would be great, thanks!

src/ansys/dpf/core/support.py
PProfizi: Ok let's create an issue there too then.

src/ansys/dpf/core/support.py
PProfizi: `self.get_type` could return an actual type/class, no? Then you could ask that class to return a deep_copy and the chain of IFs here would not be necessary.

PProfizi: Anyway, I guess this is again work to be evaluated in a different issue.

src/ansys/dpf/core/any.py
rafacanton: Issue created: https://github.com/ansys/pydpf-core/issues/3158

src/ansys/dpf/core/support.py
rafacanton: Issue here: https://github.com/ansys/pydpf-core/issues/3159

src/ansys/dpf/core/support.py
rafacanton: I joined it in the previous one: https://github.com/ansys/pydpf-core/issues/3159

PProfizi (approved):

PR #3160, opened 2026-04-29
doc: Add tutorial to understand cyclic models

rafacanton:


PProfizi:
@rafacanton sorry about that it looks like there is currently an issue with the PyAnsys action for deploying the doc of a PR. You can ignore that.

doc/sphinx_gallery_tutorials/import_data/cyclic_expansion.py
```diff
@@ -0,0 +1,248 @@
...
+=======================================
+
+:bdg-mapdl:`MAPDL`
+
+Learn how DPF handles cyclic symmetry data and the different cyclic expansion modes.
+
+When a model possesses cyclic symmetry, only one sector (the base sector) is solved.
+DPF can read these results in different ways depending on the ``read_cyclic`` option
+passed to a result operator. This tutorial demonstrates the four modes available
+(``0``, ``1``, ``2``, ``3``) and shows how to perform phase sweeping on the expanded
```
PProfizi, line 37: @rafacanton did you choose to showcase a phase sweep because this is what people usually do?
I fear that by focusing on phase sweeps we may loose the generic topic of dealing with cyclic symmetries.
Performing phase sweeps could be in another tutorial.
Or are we performing one simply because the only cyclic example we have is this one and it requires a phase sweep to show the expanded displacement field?

PProfizi: I would have preferred to showcase how to extract expanded results and use them along with expanded/non-expanded meshes.

PProfizi: What I mean is that this is very harmonic-frequency-analysis-oriented while the goal of the tutorial is to show how to deal with cyclic models in general.

doc/sphinx_gallery_tutorials/import_data/cyclic_expansion.py
```diff
@@ -0,0 +1,248 @@
...
+###############################################################################
+# Explore the cyclic metadata
+# ----------------------------
+#
+# A cyclic model exposes a
+# :class:`CyclicSupport<ansys.dpf.core.cyclic_support.CyclicSupport>` that
+# provides the number of sectors, node counts on the base sector, and the
+# mapping between low and high boundary nodes.
+#
+# The |TimeFreqSupport| describes how solution sets map to harmonic indices and
```
PProfizi, line 69: This too is specific to frequency domain analyses, but we are presenting the CyclicSupport.

doc/sphinx_gallery_tutorials/import_data/cyclic_expansion.py
```diff
@@ -0,0 +1,248 @@
...
+# only the imaginary part of the frequency is printed. Each LoadStep is a harmonic
+# index and each substep represents a cyclic mode for each one of them. For both
+# harmonic indices 0 and 4, modes are standalone, whereas for the intermediate
+# indices modes come in pairs.
+tfs = my_model.metadata.time_freq_support
+print(tfs)
+
+###############################################################################
+# Get the CyclicSupport from the ResultInfo. The CyclicSupport allows to understand
+# the cyclic information of the model. We can see that we have only 1 stage with 8
```
PProfizi, line 87: Do we have any example with several stages?
If the goal of the tutorial is to present the CyclicSupport, which happens to store information about stages, maybe we need a better example.

doc/sphinx_gallery_tutorials/import_data/cyclic_expansion.py
```diff
@@ -0,0 +1,248 @@
...
+# The following comparison shows how the number of entities grows with each
+# expansion mode.
+
+print(f"read_cyclic=0, first field entities: {len(disp_rc0[0])}")
+print(f"read_cyclic=1, first field entities: {len(disp_rc1[0])}")
+print(f"read_cyclic=2, first field entities: {len(disp_rc2[0])}")
+print(f"read_cyclic=3, first field entities: {len(disp_rc3[0])}")
+
+###############################################################################
+# Phase sweeping on expanded results
```
PProfizi, line 187: Ok so it's not that this is not great and I don't want it here, it's more that I think it deserves its own tutorial so that people can find this information better.

PR #3184, opened 2026-05-11
feat: deprecate legacy plotter

### LINKED ISSUE #2918: Deprecate the legacy `ansys.dpf.core.plotter.Plotter` class.
Remove the `Plotter` class. I guess we can just move the methods to the `DpfPlotter` class and update the method calls in relevant modules. What do you think @PProfizi ?

moe-ad:
Closes https://github.com/ansys/pydpf-core/issues/2918.

This PR implements changes required to deprecate the legacy [`Plotter`](https://github.com/ansys/pydpf-core/blob/a2407f4fd761d98272e94acd0b53ca3119173b31/src/ansys/dpf/core/plotter.py#L1627) class. By far the major blocker that has been preventing its deprecation is the [`plot_contour`](https://github.com/ansys/pydpf-core/blob/a2407f4fd761d98272e94acd0b53ca3119173b31/src/ansys/dpf/core/plotter.py#L1736) method, which is used directly by the [`Field`](https://github.com/ansys/pydpf-core/blob/ba657d6025403292f1edbfc7045c553504590ade/src/ansys/dpf/core/field.py#L527) and [`MeshedRegion`](https://github.com/ansys/pydpf-core/blob/ba657d6025403292f1edbfc7045c553504590ade/src/ansys/dpf/core/meshed_region.py#L603) classes. The `plot_contour` method does too much for a single method, which unfortunately complicates things.

In summary, this PR:
  - Breaks the monolithic logic in `plot_contour` across:
    - Smaller and more manageable module-level private functions (there is a [`_map_field_to_mesh`](https://github.com/ansys/pydpf-core/blob/ba657d6025403292f1edbfc7045c553504590ade/src/ansys/dpf/core/vtk_helper.py#L791) function whose logic overlaps some sections of `plot_contour`, so I used this opportunity to refactor things).
    - Private methods in relevant classes. We can discuss whether or not these methods should be make public.
  - Moves `Plotter.plot_chart` to a module level function (another alternative is to make it a method of [`DpfPlotter`](https://github.com/ansys/pydpf-core/blob/ba657d6025403292f1edbfc7045c553504590ade/src/ansys/dpf/core/plotter.py#L1230)).
  - Add deprecation warning to the `Plotter` class so users are informed of the eventual deprecation, and to the API documentation (on the assumption that this ships with 0.17.0).
  - Add relevant tests and updates existing tests to accommodate changes made.
  
In detail, this PR:
  - adds `DpfPlotter.add_fields_container()` as the new public method for plotting a `FieldsContainer` as a single contour.
  - splits the legacy `Plotter.plot_contour()` workflow into smaller private helpers in `plotter.py`, and these private helpers are then reused when needed.
  - moves field-to-mesh scattering logic into `MeshedRegion._scatter_field_to_location()`, and reuses that logic from `vtk_helper._map_field_to_mesh()`.
  - adds `FieldsContainer._normalize_shell_layers()` to contain the logic for normalizing multi-layer shell results before plotting.
  - updates `Field.plot()`, `FieldsContainer.plot()`, and `MeshedRegion.plot()` to use `DpfPlotter` directly instead of relying on the legacy `Plotter`.
  - moves chart plotting to the module-level `plot_chart()` function, while keeping Plotter.plot_chart() as a deprecated wrapper for compatibility.
  - keeps `Plotter` available for now, but adds deprecation warnings and deprecation notes (to the API documentation) to guide users toward `DpfPlotter`.
  - adds and updates tests to cover the new plotting code execution paths, shell-layer normalization, mesh scattering, chart plotting, and deprecated compatibility behavior.
  
  #### End results
| Plots still using former `Plotter` | After changes in this PR |
|----------|----------|
| [Example 1](https://dpf.docs.pyansys.com/version/dev/examples/00-basic/00-basic_example.html) | [Example 1 Updated](https://dpf.docs.pyansys.com/pull/3184/examples/00-basic/00-basic_example.html) |
| [Example 2](https://dpf.docs.pyansys.com/version/dev/tutorials/mapping/mapping_solid_to_skin.html#map-elemental-stress-to-the-skin-mesh) (compare several plots starting from the linked section) | [Example 2 Updated](https://dpf.docs.pyansys.com/pull/3184/tutorials/mapping/mapping_solid_to_skin.html#map-elemental-stress-to-the-skin-mesh) |

Check [here](https://dpf.docs.pyansys.com/pull/3184/api/ansys/dpf/core/plotter/Plotter.html#ansys.dpf.core.plotter.Plotter) for appearance of API deprecation notes.


jorgepiloto:
Thanks for the refactor, @moe-ad. I noticed a behavioral regression in `Plotter.plot_contour()`:

`Plotter.__init__`  still stores constructor kwargs like `off_screen`  and `notebook`. However, `plot_contour()` now instantiates a new  `DpfPlotter` from method `kwargs` only so constructor-level options are no longer used unless repeated per call.

Is this intended? Should we preserve legacy behavior by reusing the existing internal plotter/config from  `Plotter.__init__`  (or by merging stored `__init__` `kwargs` into  `plot_contour()`  `kwargs` with call `kwargs` taking precedence)?

This seems important for backward compatibility during deprecation.

moe-ad:
> Thanks for the refactor, @moe-ad. I noticed a behavioral regression in `Plotter.plot_contour()`:
> 
> `Plotter.__init__`  still stores constructor kwargs like `off_screen`  and `notebook`. However, `plot_contour()` now instantiates a new  `DpfPlotter` from method `kwargs` only so constructor-level options are no longer used unless repeated per call.
> 
> Is this intended? Should we preserve legacy behavior by reusing the existing internal plotter/config from  `Plotter.__init__`  (or by merging stored `__init__` `kwargs` into  `plot_contour()`  `kwargs` with call `kwargs` taking precedence)?
> 
> This seems important for backward compatibility during deprecation.

Great catch! The plotter created in `Plotter__init__` should be reused in `plot_contour`. I have added https://github.com/ansys/pydpf-core/pull/3184/commits/c4d256dab281e8663132809690e3cb47e34b2f73 to address that.

src/ansys/dpf/core/fields_container.py
```diff
@@ -524,6 +529,85 @@ def get_time_scoping(self):
...
         return self.get_label_scoping("time")
 
+    def _normalize_shell_layers(
+        self,
+        shell_layer: shell_layers = shell_layers.top,
+    ) -> FieldsContainer:
+        """Return a container with shell layers normalized to a single layer.
+
+        Fields whose shell_layers is ``topbottom`` or ``topbottommid`` are converted to ``shell_layer``
+        ``change_shell_layers`` operator. Fields that already have a single-layer format are passed
```
PProfizi: @moe-ad there seems to be a word missing

src/ansys/dpf/core/fields_container.py
```diff
@@ -544,18 +628,34 @@ def plot(self, label_space: dict = None, **kwargs):
...
-        merge_op = dpf.core.operators.utility.merge_fields()
-        for i, f in enumerate(fields):
-            merge_op.connect(i, f)
-        merged_field = merge_op.eval()
-        plt.add_field(field=merged_field, **kwargs)
+
+        # Build a FieldsContainer restricted to the selected label_space so we can
+        # scatter-merge onto the mesh (later fields overwrite earlier ones).
+        filtered_fc = FieldsContainer(server=self._server)
+        for label in self.labels:
```
PProfizi, line 638: If you are reusing the labels from the initial FieldsContainer, then in theory you should end-up with the same number of fields, no? I am not clear on why thsi is necessary. Or did you mean to use the label_space input of the function?

src/ansys/dpf/core/meshed_region.py
```diff
@@ -636,19 +639,31 @@ def plot(
...
-                shell_layers,
-                show_axes=kwargs.pop("show_axes", True),
+            from ansys.dpf.core import errors as dpf_errors
+            from ansys.dpf.core.common import shell_layers as eshell_layers
+            from ansys.dpf.core.fields_container import FieldsContainer
+
+            if self.is_empty():
+                raise dpf_errors.EmptyMeshPlottingError
+            pl = DpfPlotter(**kwargs)
+            shell_layer = shell_layers if shell_layers is not None else eshell_layers.top
```
PProfizi: Isn't eshell_layers.top already the default for Plotter methods such as add_field or add_fields_container?

src/ansys/dpf/core/meshed_region.py
```diff
@@ -808,3 +823,113 @@ def get_elemental_nodal_size_list(self) -> np.ndarray:
...
+            fields = [source]
+
+        # Locate the first non-empty field for metadata inference
+        reference_field = None
+        for f in fields:
+            if len(f.data) != 0:
+                reference_field = f
+                break
+        if reference_field is None:
+            raise ValueError("Cannot scatter empty source: no field contains any data.")
```
PProfizi:
```suggestion
            raise ValueError("Cannot scatter empty source: all fields are empty.")
```
Let's not use a double negative.

PProfizi (approved):

src/ansys/dpf/core/fields_container.py
moe-ad: This is just to create a new FieldsContainer with the same labels, otherwise line 644 would fail with `DPFServerException: the query is not correctly asked i.e. your collection doesn't have any LabelSpace.`

rafacanton:
@moe-ad Hi, this PR has broken some tests in pdpf-post: https://github.com/ansys/pydpf-core/actions/runs/33763061857/job/100673937720?pr=3426#step:20:1141. Can you please take a look? In this time of the release we would like to have maximum stability accross pipelines to ensure that we catch any issue... Thanks!

PR #3214, opened 2026-05-21
feat: add namespaces and keys to operator Markdown doc metadata

PProfizi:
## Summary

Adds `namespaces` to the YAML front-matter of every router-operator Markdown page generated by `generate_operators_doc.py`, so downstream tooling can filter operator pages by supported file formats without parsing the body text.

## Changes

- `generate_operators_doc.py`: store `supported_file_types[namespace]` as a sorted list instead of a pre-joined string.
- `operator_doc_template.j2`:
  - Emit a `namespaces` block in the YAML front-matter for router operators (e.g. `acceleration.md` gets `hdf5: [h5dpf]`, `lsdyna: [d3plot, ...]`, `mapdl: [rst, ...]`).
  - Fix the body list rendering with `| join(", ")` - required because the value is now a list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ MANIFEST
 *.manifest
 *.spec
 
+# Bandit
+info_bandit.txt
+
 # Installer logs
 pip-log.txt
 pip-delete-this-directory.txt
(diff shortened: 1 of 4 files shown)


AnsMelanie (approved):
@PProfizi 

This looks good to me as you have added the namespace/keys metadata that were missing.

PR #3233, opened 2026-05-29
feat: report full calendar version from server.version

PProfizi:
Use DataProcessing_getServerVersionFull / DataProcessing_getServerVersionFull_on_client to expose the full calendar-like version string (e.g. '2027.1.0pre0') from all server types.

A guard on major >= CALENDAR_VERSIONING_FIRST_MAJOR (2027) ensures backward compatibility with older DPF servers that only support the legacy major.minor API; those continue to return a 'X.Y' string.

Changes:
- _version.py: add CALENDAR_VERSIONING_FIRST_MAJOR = 2027 constant
- GrpcServer.version: first calls get_server_version_on_client; if major >= 2027 calls get_server_version_full_on_client for micro+modifier components
- InProcessServer.version: same pattern via get_server_version / get_server_version_full (direct CAPI, not the grpc_client variants which require the gRPC client layer to be initialized)
- BaseService._get_server_info: same guard so LegacyGrpcServer.version (which reads from info['server_version']) also reports the full version
- DataProcessingGRPCAPI: add data_processing_get_server_version_full_on_client reading microVersion and modifierVersion from the gRPC ServerInfoResponse proto
- test_checkversion.py: add test_get_server_version_format and test_server_info_version_format covering all three server types

### DIFF
--- a/src/ansys/dpf/core/_version.py
+++ b/src/ansys/dpf/core/_version.py
@@ -48,3 +48,7 @@ class ServerToAnsysVersion:
 
 
 server_to_ansys_version = ServerToAnsysVersion()
+
+# First DPF major version using calendar versioning (YEAR.REVISION.MICRO.MODIFIER).
+# Servers with major >= this support DataProcessing_getServerVersionFull.
+CALENDAR_VERSIONING_FIRST_MAJOR = 2027
--- a/src/ansys/dpf/core/core.py
+++ b/src/ansys/dpf/core/core.py
@@ -34,6 +34,7 @@ from ansys.dpf.core import errors, server as server_module
 
 if TYPE_CHECKING:  # pragma: noqa
     from ansys.dpf.core import AnyServerType
+from ansys.dpf.core._version import CALENDAR_VERSIONING_FIRST_MAJOR
 from ansys.dpf.core.check_version import server_meet_version, version_requires
 from ansys.dpf.core.runtime_config import (
     RuntimeCoreConfig,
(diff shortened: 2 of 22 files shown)


src/ansys/dpf/core/_version.py
```diff
@@ -48,3 +48,7 @@ def __getitem__(self, item):
 
 
 server_to_ansys_version = ServerToAnsysVersion()
+
+# First DPF major version using calendar versioning (YEAR.REVISION.MICRO.MODIFIER).
```
BClappe, line 52: More or less Calendar versioning, but this is a minor thing.

src/ansys/dpf/core/core.py
```diff
@@ -613,13 +614,41 @@ def _get_server_info(self):
...
+                    micro=serv_ver_micro,
+                    modifier=serv_ver_modifier,
+                )
+            version_str = (
+                str(int(serv_ver_maj))
+                + "."
+                + str(int(serv_ver_min))
+                + "."
+                + str(int(serv_ver_micro))
+                + str(serv_ver_modifier)
```
BClappe, line 648: Should it be glued together (27.1.0pre0) or should a dash separate them (27.1.0-pre0)?
I would opt-in for the dash separation.

src/ansys/dpf/core/server_types.py
```diff
@@ -974,7 +979,17 @@ def version(self):
...
             api.data_processing_get_server_version_on_client(self.client, major, minor)
-            self._version = str(int(major)) + "." + str(int(minor))
+            if int(major) >= CALENDAR_VERSIONING_FIRST_MAJOR:
+                micro = integral_types.MutableInt32()
+                modifier = integral_types.MutableString(size=0)
+                api.data_processing_get_server_version_full_on_client(
+                    self.client, major, minor, micro, modifier
+                )
+                self._version = (
+                    str(int(major)) + "." + str(int(minor)) + "." + str(int(micro)) + str(modifier)
```
BClappe, line 989: Should there be a dash separating micro and modifier?

BClappe (approved):

rafacanton:
@PProfizi Thanks! I guess that we can also modify the source code and tests to instead of check `meet_server_version(12.0)` to `meet_server_version(2027.1.0pre0)`. Thanks!

src/ansys/dpf/core/core.py
rafacanton: I think it makes sense

PProfizi: https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-releases
Python's packaging guidelines say it should be glued.

It is the [Semver](https://semver.org/) convention which mentions using a hyphen, and we are moving away from SemVer, so I'd rather follow Python's best practices here, as even though we are discussing DPF's version, it will be packaged as a standalone Python library.
The question really is: how does DPF do it?

rafacanton (approved):

PR #3275, opened 2026-06-22
fix: broken link in documentarian section

moe-ad:
As the title states. Link broken [here](https://dpf.docs.pyansys.com/version/stable/contribute/documentarian.html#contributing-to-the-documentation):
image

The fix is to change the location of the include statement, it shouldn't come between a label and the target section title.


moe-ad:
I will merge this @PProfizi. I tested locally and the deployed PR docs also confirm the fix.

PR #3337, opened 2026-07-22
fix: Operator doc generation for operator pins using uint64 

PProfizi:
To prepare the server update bringing support for uint64 as a pin data type.

### DIFF
--- a/src/ansys/dpf/core/mapping_types.py
+++ b/src/ansys/dpf/core/mapping_types.py
@@ -78,6 +78,8 @@ map_types_to_python["vector<bool>"] = "list[bool]"
 map_types_to_python["vector<int32>"] = "list[int]"
 map_types_to_python["vector<double>"] = "list[float]"
 map_types_to_python["vector<string>"] = "list[str]"
+map_types_to_python["vector<uint64>"] = "list[int]"
+map_types_to_python["uint64"] = "int"
 map_types_to_python["b"] = "bool"
 
 
@@ -104,6 +106,8 @@ def reflection_type_to_cpp_type(reflection_type: str) -> str:
         reflection_type = "streams"
     if reflection_type == "int32":
         return "int"
+    elif reflection_type == "uint64":
+        return "uint64_t"
     elif reflection_type in ["double", "bool"]:
         return reflection_type
     elif reflection_type == "string":
--- a/.github/workflows/update_operators.yml
+++ b/.github/workflows/update_operators.yml
@@ -113,6 +113,7 @@ jobs:
         run: |
           pip install -U pip
           pip install --group build
+          pip install --group operator_doc
 
       - name: "List installed packages"
         shell: bash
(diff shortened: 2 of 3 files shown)


src/ansys/dpf/core/mapping_types.py
```diff
@@ -78,6 +78,8 @@ def __missing__(self, key):
 map_types_to_python["vector<int32>"] = "list[int]"
 map_types_to_python["vector<double>"] = "list[float]"
 map_types_to_python["vector<string>"] = "list[str]"
+map_types_to_python["vector<uint64>"] = "list[int]"
```
rafacanton, line 81: @PProfizi is it an issue that we use `int` for both 32 and 64 integers? Does the Python int cover both?

PProfizi:
@jvidalor this fixes the update pipeline as you can see [here](https://github.com/ansys/pydpf-core/actions/runs/29932486581) and [here](https://github.com/ansys/pydpf-core/pull/3336)

src/ansys/dpf/core/mapping_types.py
PProfizi: @rafacanton there is not distinction in basic Python types. I quote Copilot:

Reasons:

Python int is arbitrary-precision, so it fully represents the entire uint64 range (0 to 2^64-1) with no overflow. There's no native unsigned integer type in Python.
These mappings feed generated operator docstrings and type annotations. The annotation must be a plain Python type importable via find_class_origin from ansys.dpf.core. numpy.uint64 wouldn't fit that mechanism, and there's no Uint64 class in the package.
Existing precedent: int32 already maps to int, so mapping uint64 to int too is consistent — the connect/get APIs accept/return Python int regardless of the underlying C++ width.
The signedness/width distinction only matters at the C++ boundary, which is already handled separately by reflection_type_to_cpp_type returning uint64_t. So int on the Python side is the best fit.

rafacanton:
To actually fully support uint, I was expecting changing this enum: https://github.com/ansys/pydpf-core/blob/50847c4014b6472ca87005011c532520201ba09e/src/ansys/dpf/core/common.py#L152-L153

jvidalor:
> To actually fully support uint, I was expecting changing this enum:
> 
> https://github.com/ansys/pydpf-core/blob/50847c4014b6472ca87005011c532520201ba09e/src/ansys/dpf/core/common.py#L152-L153

@PProfizi ok, the error that we see in the pipelines are related to a fluids error. Please let me know once you address Rafa's comment (we may need to add a test to check it with an operator).

PProfizi:
> To actually fully support uint, I was expecting changing this enum:
> 
> https://github.com/ansys/pydpf-core/blob/50847c4014b6472ca87005011c532520201ba09e/src/ansys/dpf/core/common.py#L152-L153

Ok but I'll need the update from the server to be able to properly expose them. So let's do this in two steps. I merge the fix so the update can pass, and then I do another PR to properly expose the new uint server APIs.

rafacanton (approved):

PR #3381, opened 2026-08-11
comment tests before updating the code 

MichaelNale:
Comments test that will fail with TFS PR updating dimension and num component management


jvidalor (approved):
Approving, since this issue will be fixed soon

PR #3388, opened 2026-08-13
comment doc for field def update

MichaelNale:
comment documentation field def examplebefore merging framewor k pr concerning dimension


jvidalor (approved):
Please create an issue for all these commented tests (also from yesterday)

PR #3390, opened 2026-08-13
Fix set_config_option issue

Sangsimon:
- op.config.set_config_option(config_name="permissive", config_value=True) doesn't work since the op.config returns a temp obj whose state is discarded when the line ends. So set_config_option is ineffective.
- The workaround is to create a temp object,  modify it and call the setter: cfg = op.config cfg.set_config_option(config_name="permissive", config_value=True) op.config = cfg
- Since this sol is not user friends, a _ConfigProxy class is introduced that does this technique underneath

- Operator.config is modified so that the call goes through a proxy object and is handled by implementing the above solution


PProfizi (approved):
I'm ok with the principle of the fix, thanks @Sangsimon!

MichaelNale (approved):

PR #3398, opened 2026-08-20
feat: entry points

### LINKED ISSUE #2946: Expose scripts in .ci folder via entry points in pyproject.toml
Suggested by @PProfizi.

1. We would like to get rid of the `.ci` directory from the project root (see #2927 and #2940 also), by moving these scripts to some other location.
2. At the same time, python scripts in `.ci` are useful for local development workflows, most especially ensuring that operators' documentation are synced with a particular server version. That implies easing how users can use these scripts locally when the need arises.

We can achieve both by moving these scripts to a module in one of the existing subpackages in `ansys.dpf.core` (or creating a new subpackage?), and then exposing the invocation of these scripts through entry points so that they can then be run intuitively via the CLI.

@PProfizi please add any other point I might have missed.

moe-ad:
Closes #2946.

Most scripts in `.ci/` have been removed / moved as the case may be (see #2946 for more context). The only ones left are:
  - build_wheel.py
  - run_examples.py
  - update_dpf_dependencies.py

Additionally, relevant scripts have been exposed as entry points and also through tox.


pyproject.toml
```diff
@@ -137,6 +138,8 @@ Tracker = "https://github.com/ansys/pydpf-core/issues"
 
 [project.scripts]
 dpf_generate_operators_doc = "ansys.dpf.core.documentation.generate_operators_doc:run_with_args"
+dpf_build_operators_html_doc = "ansys.dpf.core.documentation.build_operators_html_doc:run_with_args"
```
PProfizi: @moe-ad could we maybe differentiate this one from the one above by calling it `dpf_generate_operators_doc_legacy`? or at least `dpf_generate_operators_doc_html` to have naming consistency.

tox.ini
```diff
@@ -16,6 +16,13 @@
 # For html documentation generation, run `tox -e doc-html`. To clean previously existing documentation before generating
 # a new one, run `tox -e doc-clean,doc-html`
 
+# To generate the legacy static HTML operator documentation page for the DPF install detected on the machine,
+# run `tox -e build-operators-html-doc`.
+
+# To regenerate the operator modules from the DPF install detected on the machine, run `tox -e generate-operators`.
```
PProfizi, line 22:
```suggestion
# To regenerate the operator modules from the DPF install detected on the machine, run `tox -e dpf_generate_operators`.
```
Shouldn't we make all these environment names consistent with the entry-points? Or is it not usually done this way?

PProfizi: Ok that's probably not necessary.

PProfizi (approved):

pyproject.toml
moe-ad: Done!

PProfizi (approved):

Issue #3421, opened 2026-09-01
Adopt lockfiles in pydpf-core


moe-ad:
We want to start using lockfiles for (the list is not exhaustive):
- more reproducible installs across environments
- better supply chain security
- conforming with current best practice

Among current tooling with support for lockfiles, the obvious choices are `uv` and `poetry` and we have more or less decided to go with `uv`.

PR #3445, opened 2026-09-09
feat: switch build-backend

moe-ad:
WIP. Related to #3421.


jorgepiloto (approved):
Nice discovery! Thanks for migrating to Hatch. It simplified lots of things.

.gitignore
```diff
@@ -193,14 +193,6 @@ dpf-standalone/
 
 search/
 
-# ansys-dpf-gate, ansys-grpc-dpf and ansys-dpg-gatebin
```
PProfizi, line 196: @moe-ad why do you need to remove this from the gitignore?

PR #3189, opened 2026-05-12
Fix xml config with xml path

cbellot000:
This PR fixes the server config: when an xml was provided, it was never given to the server

PR #3472, opened 2026-09-18
Fix local server launch with custom XML context paths containing spaces

bvdbr:
This PR fixes local DPF server startup on Windows when a custom XML licensing context is located in a path containing spaces.

Previously, the launcher assembled the process command as a plain string. As a result, Windows split context paths such as `C:\Program Files\...` into multiple arguments, causing the server to receive an invalid `--context` value and fail to start. The launcher now builds local server commands as structured argument lists


rafacanton:
Thanks @bvdbr for this. Do we know why this started failing several weeks ago? Or this issue was always there? I find it odd because the usual place to install Ansys is `C:\Program Files\ANSYS Inc\...`, so it should have been caught in tests running with the Unified Install by @laurent-ge

bvdbr:
@rafacanton The PR #3189 merged on September 4th introduced the issue : 

`run_cmd.append(f"--context {context.xml_path}")`

This line passes the path without proper quoting. On Windows, a path beginning with` C:\Program Files\...` was therefore split into multiple arguments.

rafacanton (approved):

PR #3474, opened 2026-09-18
docs: fix 3 grammar errors in comments and documentation

Avicennasis:
## Summary

Fix 3 grammar errors across 2 files.

`src/ansys/dpf/core/operators/logic/component_selector_fc.py`
- `one or several component index` → `one or several component indices`

`src/ansys/dpf/core/operators/math/component_wise_product.py`
- `of same\ndimensionality` → `of the same\ndimensionality`
- `of same` → `of the same`

Otherwise comments, documentation and message strings only — no functional changes.


rafacanton:
@moe-ad Code in `‎src/ansys/dpf/core/operators/` is auto-generated, its source of truth is not here. If we modify it, it is futile

Avicennasis:
You're completely right, @rafacanton — thanks for catching this, and apologies for the noise.

Both files this PR touches are under `src/ansys/dpf/core/operators/`, and the header of each says so plainly:

```python
"""
component_wise_product

Autogenerated DPF operator classes.
"""
```

I should have read that before filing. The directory even ships `operator.mustache` and `build.py` alongside. Any change here gets overwritten the next time the operators are regenerated, so the PR is exactly as futile as you say.

Closing it. The underlying nits were trivial (`of same` → `of the same`, `component index` → `component indices` in the docstrings), and since they originate from the operator specifications rather than this repo, they're not worth chasing further from the outside — but flagging in case the wording is worth correcting at the source.

I've also fixed the gap on my end: I now check every file's leading lines for generated-code markers before proposing any edit, which would have caught this one immediately.

